PR #7, opened 2020-09-08
Update requirements.txt

Faithy3:
add missing dependencies

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,5 +7,5 @@ orjson
 aiofiles
 oppai
 uvloop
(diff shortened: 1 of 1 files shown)


cmyui (changes requested):
Hey! Thanks for the PR I kept forgetting to add these haha. One thing though, would you be able to change "mysql-connector" to "mysql-connector-python"? mysql-connector is an inactive branch, more info on that [here](https://github.com/pypa/warehouse/issues/5537).

Cheers!

PR #11, opened 2020-10-09
Add build-essential to instructions

PoltixeTheDerg:
@Faithy3 asked me to make a PR for this, as he said it's required to build oppai


Faithy3:
At least make and gcc were required, I installed build-essential as a catch-all after the second dependency error.

PR #19, opened 2020-10-17
Certgen & QOL changes

jacksonisiah:
Up to you if you want to merge this. Certgen was working last time I used it which was about a month ago :)


cmyui:
Hey, there seems to be some sort of difference in the format from this cert to akatsuki/ripple/etc.. i think this may be the same as the ainu certificate, which seems to work.. sometimes? lol

once i figure out the differences, i might add a certgen (but i'm also not sure about how the osu team will feel about it xd)

cmyui:
one point for that pr if you go through with it - `ppy.sh` should become a variable now that we have the `-devserver` flag in the osu! stable client

7mochi:
> Hey, it's been a while, but if you'd like to PR the certgen into `/tools` I'd be willing to merge it - it's pretty useful for those who want to run a server locally.
> 
> (Generally we suggest prod users to use cloudflare flex to sidestep tls entirely on the server-side)

Wouldn't it be better to use mkcert for localhost certs?

tsunyoku:
> > Hey, it's been a while, but if you'd like to PR the certgen into `/tools` I'd be willing to merge it - it's pretty useful for those who want to run a server locally.
> > (Generally we suggest prod users to use cloudflare flex to sidestep tls entirely on the server-side)
> 
> Wouldn't it be better to use mkcert for localhost certs?

+

cmyui:
> > Hey, it's been a while, but if you'd like to PR the certgen into `/tools` I'd be willing to merge it - it's pretty useful for those who want to run a server locally.
> > (Generally we suggest prod users to use cloudflare flex to sidestep tls entirely on the server-side)
> 
> Wouldn't it be better to use mkcert for localhost certs?

Perhaps - I had some issues with generating one for local dev (perhaps an issue with my usage), and have always fallen back to using this directly. If you'd like to propose a PR for it, I'm open to the best suggestion!

cmyui:
Might also be useful to include a small document to define the process of setting up a local server, including hosts file redirection and such.

7mochi:
I was thinking of adding it also in https://github.com/osuAkatsuki/bancho.py/pull/320 but wouldn't the README be too bloated? I think the best thing would be to move some of those instructions to [github wiki](https://docs.github.com/en/communities/documenting-your-project-with-wikis/about-wikis)

NiceAesth:
I've spoken with Jackson and he is open to fixing this PR up. (I think documentation should be a separate PR btw since translations and such)

cmyui:
> I was thinking of adding it also in https://github.com/osuAkatsuki/bancho.py/pull/320 but wouldn't the README be too bloated?

Yes, keep PRs small and specific, this PR is already a bit large for my liking (e.g. I think we could have split up the ssl removal & and the docker setup, and reviewed those separately).

Big PRs are much more difficult to keep up to date while master is worked on, and also harder to review.

minisbett:
I would disagree with renaming ext to external

PR #28, opened 2020-10-20
Fix SQL column name issue when adding Channel to ChannelList()

Yo-ru:


cmyui:
Hey, thanks for the pr, but unfortunately i already had this one fixed on my local branch but had forgotten to push xD, i would accept it otherwise..

i simply re-added `*args, **kwargs` to the channel class (and player as well since it had a similar thing) to keep the query clean B)

PR #32, opened 2020-11-09
Tiny fix: correct byte length

Pure-Peace:
When trying to write non-English characters to the packets, it will get an incorrect length.
Incorrect length will cause the osu! client to not display some packets normally (such as notifications)

len("h") -> 1 √
len("嗨") -> 1 ×
len("嗨".encode()) -> 3 √


cmyui:
Thanks for the fix :D

PR #35, opened 2020-11-13
Mania PP

Yo-ru:
Pushed my changes then realized I fucked up, reverted them, pushed actual changes and updated requirements.


Yo-ru:
This is a fully functional example of mania pp, but there are probably cleaner and faster ways to do this as [cmyui](https://github.com/cmyui) has pointed out. So this wont be merged.

PR #46, opened 2021-01-02
fix no show '!' on !help

comfyy:


Airkek:
image
you should take command prefix from glob.config

comfyy:
true

PR #47, opened 2021-01-03
utilize the domain config value more

Ausakura:
this pr adds the change that makes it so that the domain configuration is actually used in the domain itself rather than just for certain fields; which makes it easier for making a ui to access avatars better

i tested this on my own fork and setup and it worked fine for me personally but im not sure if problems will arise

### DIFF
--- a/domains/ava.py
+++ b/domains/ava.py
@@ -8,7 +8,7 @@ from objects.glob import config # just the config
 
 """ ava: avatar server (for both ingame & external) """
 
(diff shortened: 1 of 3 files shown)


cmyui:
Discussed in the [Discord](https://discord.gg/ShEQgUx) - planning on using nginx for the avatars & screenshots soon, and the c[e4-6]?.ppy.sh & osu.ppy.sh domains are only used through the osu! client, so domains are not necessary (not planning on supporting custom clients in the official repo).

PR #50, opened 2021-02-13
Moved Bancho domain regex pattern to regexes.py and cleaned !switchserv

memsdm05:
### Bancho Regex Refactor
I think there can be applications where the Bancho domain regex can be used. Previously, the only location that this was located in was the [domain object initialization in cho.py](https://github.com/cmyui/gulag/blob/master/domains/cho.py#L40). While it works as an ad-hoc solution, we should consider future implementations that can use this regex. Therefore, I moved this pattern into constants/regexes.py so it can be referenced by the project.

### !switchserv modifications
osu!client's current implementation of the redirect cho packet is only limited to valid .ppy.sh Bancho subdomains. 

https://github.com/cmyui/gulag/blob/8febf963a9234994db3aa777e9e01383dcb49260/constants/commands.py#L507-L509

Using the edit previously described, I now check if the domain is a valid Bancho domain. Here is me testing it.

image

The redirect notif popped up and I seemed to be reconnected.

### DIFF
--- a/constants/commands.py
+++ b/constants/commands.py
@@ -513,8 +513,11 @@ async def switchserv(p: 'Player', c: Messageable, msg: Sequence[str]) -> str:
     if len(msg) != 1:
         return 'Invalid syntax: !switch <endpoint>'
 
+    if not regexes.bancho_url.match(msg[0]):
+        return 'Invalid URL: can only switch to c[e4-6].ppy.sh domains'
+
     p.enqueue(packets.switchTournamentServer(msg[0]))
-    return 'Have a nice journey..'
+    return f'Switching to {msg}... Have a nice journey!'
 
 # rest in peace rtx - oct 2020 :candle:
 #@command(Privileges.Dangerous, hidden=True)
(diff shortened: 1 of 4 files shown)


constants/commands.py
```diff
@@ -513,8 +513,11 @@ async def switchserv(p: 'Player', c: Messageable, msg: Sequence[str]) -> str:
     if len(msg) != 1:
         return 'Invalid syntax: !switch <endpoint>'
 
+    if not regexes.bancho_url.match(msg[0]):
+        return 'Invalid URL: can only switch to c[e4-6].ppy.sh domains'
+
     p.enqueue(packets.switchTournamentServer(msg[0]))
-    return 'Have a nice journey..'
+    return f'Switching to {msg}... Have a nice journey!'
```
memsdm05, line 520: crap i meant to do msg[0] not msg

cmyui:
Hey, most of the point of the command is to be able to switch to alternate servers, for example `!switchserv c.akatsuki.pw`, so this isn't quite the right regex for the job.. perhaps it could be checked with another new regex, although I don't think this command will ever be used enough to warrant any such checks :P

i do see u learnin the server structure tho 😳😳

PR #62, opened 2021-03-12
Upgrade Channels

memsdm05:
### Main Addition
Channels now use id_name for identification instead of _name and the name property. This opens up the possibility for more esoteric channels (i.e. clan channels). 

### Extras
- When changing to a new team or when a match switches to a team mode, teams are randomly assigned blue or red (coinfip)
- Player channel list is now a set

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2020 cmyui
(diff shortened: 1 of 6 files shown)


memsdm05:
convo with @cmyui about instance_ids and stuff + merge is screwy. Better push in the works

PR #75, opened 2021-03-17
Rewrote recalc command, added all scores recalc

alowave223:
Tested and all working fine for me, also logging all possible errors (should it logged only with debug is on?).


alowave223:
Oof, forgot about check valid mode combinations

alowave223:
Also i think there should be filter by mode_vn for more flexibility?

PR #77, opened 2021-03-28
No longer friends with yourself.

memsdm05:
This behavior replicates Bancho, as seen in this video.

https://streamable.com/vth5kv

Given how explicitly self-friending is mentioned in code, I believe @cmyui intentionally created this functionality. If so, I propose a follow up commit which allows toggling self-friending in config. In my own experience I find self-friending jarring, but others may have differing opinions.


memsdm05:
instead of config you can now add and remove yourself as a friend.

PR #78, opened 2021-03-29
fix: regexes.mapfile

Pure-Peace:
A very small problem,
after resolving this, grouping matching will work correctly.

[https://regexr.com/5pitg](https://regexr.com/5pitg)


cmyui:
The original reason I wrote the (?: )? around the artist was because of this map.. but in hindsight this probably SHOULD be treated as an invalid filename since the map is incomplete.

Thanks for the PR! :) I'll be looking at #79 next.

image

cmyui:
Oh also, I guess I can also remove the (?:)? around creator entirely now.

Pure-Peace:
✨🥰✨

PR #81, opened 2021-04-03
Fix chimu download

Atakku:


KotRikD:
I think you need some global flag for chimu/another api.

PR #83, opened 2021-04-10
Add auto status update for beatmaps

alowave223:
Automatic updating of the status of maps is really a necessary thing on any private server, so here is my implementation of this functionality, in short, my idea is this: to call the function to check the status of maps only when it is necessary and only when the map is not ranked/loved. Also, here I took into account the fact that the status can be changed by the !map command and made a check for this, even if one map from the whole set has a changed status, it will not change for it, and the rest of the maps from the set will change.

### DIFF
--- a/domains/osu.py
+++ b/domains/osu.py
@@ -1004,6 +1004,13 @@ async def getScores(p: 'Player', conn: Connection) -> Optional[bytes]:
                     glob.cache['unsubmitted'].add(map_md5)
                     return b'-1|false'
         else:
+            # try to update bmap status after status timeout
+            if bmap.last_check + glob.config.map_status_timeout <= int(time.time()):
+                # check if map updated then
+                # cache it with new status
+                if (updated := await bmap.update_status(bmap.md5)):
+                    bmap = updated
+
             # found in sql - add to cache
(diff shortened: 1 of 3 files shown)


alowave223:
Also there is should be sql update for maps table
`ALTER TABLE maps ADD last_check int(10);`

cmyui:
thanks for the pr and sorry for the reaaally long delay; i've had this in the back of my mind for a while now but i've decided i want to not only write this myself, but actually rewrite most of the beatmap handling code on gulag.. theres quite a bit of repetitive code and i'm sure i can make it better now, especially with my recent escapades into akatsuki's old ripple codebase..

you're very welcome to continue pring other stuff, but it's probably safest to stay away from implementing larger systems in the main classes (player/map/score/leaderboard/etc), since i'm pretty ocd about getting the implementation as cancer-free as possible lol. (this applies to pretty much anyone reading)

cheers

cmyui:
oh also, i should mention that i hadn't actually thought about this system before you submitted the pr, so if you (or anyone else reading) notices any missing subsystems like this, please open an issue about it or dm me anywhere :)

PR #85, opened 2021-04-18
Updated get_player_status api request

def750:
Added last seen timestamp when user is offline to get_player_status request.

### DIFF
--- a/domains/osu.py
+++ b/domains/osu.py
@@ -1420,7 +1420,10 @@ async def api_get_player_status(conn: Connection) -> Optional[bytes]:
     if not p:
         # no such player online
         res = await glob.db.fetch('SELECT latest_activity FROM users WHERE id = %s', [pid])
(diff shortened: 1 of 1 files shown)


domains/osu.py
```diff
@@ -1419,7 +1419,11 @@ async def api_get_player_status(conn: Connection) -> Optional[bytes]:
 
     if not p:
         # no such player online
-        return JSON({'online': False})
+        res = await glob.db.fetch('SELECT latest_activity FROM users WHERE id = %s', [pid])
+        if not res or res['latest_activity'] == 0:
+            return JSON({'online': False, 'last_seen': 'Never'})
```
cmyui, line 1424: For `res['latest_activity'] == 0`, it would be preferrable to just send 0 back rather than a string to make parsing more consistent, so that check can be removed.

As for `if not res`, it should remain consistent with the other API handlers and return `(404, b'Player not found.')`.

cmyui (changes requested):

PR #109, opened 2021-06-24
Simplify getting user's IPA

ledlamp:
x-forwarded-for is all u need, keep it simple and dont bother with the other random ones. Unless u want to prevent IPA spoofing but that will add some complexity.

Ideally it'd use the client socket address if there is no reverse proxy but i guess nobody's gonna be doing that.

### DIFF
--- a/domains/cho.py
+++ b/domains/cho.py
@@ -65,15 +65,8 @@ async def bancho_http_handler(conn: Connection) -> bytes:
 
 @domain.route('/', methods=['POST'])
 async def bancho_handler(conn: Connection) -> bytes:
-    if 'CF-Connecting-IP' in conn.headers:
-        ip = conn.headers['CF-Connecting-IP']
-    else:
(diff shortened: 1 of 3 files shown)


cmyui:
albeit a small change, i'm not sure which solution i want to go for at the moment so i'm going going to close this alongside the new duplicate #114, though i agree that we should either be using the first ip from x-forwarded-for or at least checking whether the connection is coming from one of cloudflare's ips; perhaps we can even use the socket address if they're not on cloudflare.

i'm curious as to how other http server implementations use x-forwarded-for, but one of the above solutions will be implemented.

ledlamp:
> as for on-cloudflare.. i'm not sure exactly what their servers do with x-forwarded-for
Same thing as any other typical reverse proxy. Usually the same as `CF-Connecting-IP`, however if the client is connecting to cloudflare via another proxy and X-Forwarded-For header already exists, cloudflare appends its ip address to the right of the list.

PR #112, opened 2021-07-04
Banners support

def750:
Added basic banners support to use in guweb


def750:
Closing cuz t heres unwanted commit fu..

PR #113, opened 2021-07-06
Remove get_leaderboard player didn't played that mode

Varkaria:


Varkaria:
image

because i think that's too anoying too see player didn't played that mode but still in leaderboard

PR #114, opened 2021-07-11
Remove CF-Connecting-IP for security purposes

atteure:
An exploit was recently found, mainly in Ripple stacks (but also present in servers that utilize the CF-Connecting-IP header to find and set the user's IP), that allows players to essentially spoof their IP and location using CF-Connecting-IP. This was brought up (rather helpfully) by @RealistikDash. The main issue is that if the server allows the IP to be set using said header, the player can pretty much change it and then use it freely. Servers, including ones like gulag and Asahi that do not use Cloudflare or has a server switcher, it is vulnerable.

Just thought I'd PR this; entirely up to you if you wanna take the risk.

### DIFF
--- a/domains/cho.py
+++ b/domains/cho.py
@@ -78,13 +78,9 @@ async def bancho_http_handler(conn: Connection) -> bytes:
 
 @domain.route('/', methods=['POST'])
 async def bancho_handler(conn: Connection) -> bytes:
-    if 'CF-Connecting-IP' in conn.headers:
-        ip_str = conn.headers['CF-Connecting-IP']
-    else:
-        # if the request has been forwarded, get the origin
-        forwards = conn.headers['X-Forwarded-For'].split(',')
-        if len(forwards) != 1:
-            ip_str = forwards[0]
+    forwards = conn.headers['X-Forwarded-For'].split(',')
+    if len(forwards) != 1:
+        ip_str = forwards[0]
         else:
             ip_str = conn.headers['X-Real-IP']
 
--- a/domains/osu.py
+++ b/domains/osu.py
@@ -2358,15 +2358,10 @@ async def register_account(
                 # network tab of cloudflare, so it sends the iso code.
                 country_acronym = conn.headers['CF-IPCountry']
             else:
-                # backup method, get the user's ip and
-                # do a db lookup to get their country.
-                if 'CF-Connecting-IP' in conn.headers:
-                    ip_str = conn.headers['CF-Connecting-IP']
-                else:
-                    # if the request has been forwarded, get the origin
-                    forwards = conn.headers['X-Forwarded-For'].split(',')
-                    if len(forwards) != 1:
-                        ip_str = forwards[0]
+                # if the request has been forwarded, get the origin
+                forwards = conn.headers['X-Forwarded-For'].split(',')
+                if len(forwards) != 1:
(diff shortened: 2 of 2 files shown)


tsunyoku:
The only issue provided with this, atleast as far as I know, is that there is no alternative method for grabbing the users real IP for those who **do** use Cloudflare. While it is a risk that people can spoof the IP returned, I don't think some people would like to disable Cloudflare for this purpose

jiniux:
The server should rather check if the connecting IP is from cloudflare before using that header.

tsunyoku:
If we ignore the changes this commit makes, it already does I believe.

jiniux:
Well, if the exploit was detected, then it doesn't. It should check that the IP belongs to defined subnets available on https://www.cloudflare.com/it-it/ips/

Atakku:
Just make your servers only accept connections through cloudflare if using cloudflare smh.

You can make a config option to enable CF mode, where it would use CF header and drop all connections unless they come from:
https://api.cloudflare.com/#cloudflare-ips-properties

cmyui:
yea this was discussed in the [commit where this was added](https://github.com/cmyui/gulag/commit/62a1b1922c5e740845a52b8fb47d77037d39448c) and is also a duplicate of #109; seems like we can either check for cloudflare ips or just use the first ip in x-forwarded-for. gonna hold off until i research more into how other http server implementations use x-forwarded-for before making any action.

either way this is minor concern since gulag only tracks ips for logins at the moment and doesn't use them for any security checks. if people wanted to fake their country they could just use a vpn.

ledlamp:
> gonna hold off until i research more into how other http server implementations use x-forwarded-for before making any action.

The [Express](https://expressjs.com/) framework replaces `req.ip` with the appropriate value from the `X-Forwarded-For` header, replaces `req.hostname` with the value of `X-Forwarded-Host`, and `req.protocol` with `X-Forwarded-Proto`, all IF the framework trusts the client's socket address as a proxy as configured by the `'trust proxy'` setting.

For example, with `app.set('trust proxy', '127.0.0.1')`, the framework will use those headers only if the connection is from localhost. It will use the nearest trusted level of the XFF list... soo the right-most value if there is only one hop. Such way it doesn't matter what the client connecting to i.e. nginx sets as X-Forwarded-For (which would be added to the left).

You could add the entire list of cloudflare's IPs to express, so if the connection from Cloudflare, it will recognize it in the rightmost place on the list and use the next address to the left of that to substitute into the request properties.

Other ways to configure it include setting a number of hops to trust, i.e. `2` if you have cloudflare -> nginx -> app and no other possible way around it. More information: https://expressjs.com/en/guide/behind-proxies.html

ledlamp:
I think the best thing to do is implement similar proxy trust functionality, and have the end user provide a list of trusted address blocks in their config.

So for example with a typical setup of cloudflare -> nginx -> app, the list would contain `127.0.0.1` + [cloudflare's addresses](https://www.cloudflare.com/ips/). If socket address is in list (i.e. 127.0.0.1), check last (rightmost) value of XFF, if it's also in list (i.e. [103.21.244.69](https://whatismyipaddress.com/ip/103.21.244.69)), check the previous (left) one, repeat until find one that's not in list.

This'll allow full control for the sysadmin to set it up however they want with proper security.

PR #115, opened 2021-07-11
only display ranked maps in scores api

tsunyoku:


cmyui:
perhaps this could be an option but i'd definitely still want to provide functionality for fetching unranked scores.. for example the !recent command in a discord bot that can show the most recent play, even if it's something like a failed play on an unranked map.

PR #120, opened 2021-07-26
fix max char size

tsunyoku:
unique constraint forces max byte size to 767, so we need these to have max char count of 191 for it to work :p


tsunyoku:
im gonna push this as part of other fixes im making, so ill close this

PR #122, opened 2021-07-27
Add include_loved argument to /api/get_player_scores

JKBGL:


cmyui:
this code is only used for `scope == 'best'`, and can be moved into the block.
image

PR #123, opened 2021-07-30
ubuntu 20.04 support

### LINKED ISSUE #107: ubuntu 20.04 dropped support for tls versions used by osu!stable
> - ubuntu 20.04 & nginx have unknown issues? i recommend using 18.04

apparently osu uses older tls 1.2 or 1.1 which ubuntu 20.04 doesn't support but 18.04 does.

I tried adding 
```
ssl_protocols TLSv1 TLSv1.1 TLSv1.2 TLSv1.3;
ssl_ciphers "EECDH+AESGCM:EDH+AESGCM:AES256+EECDH:AES256+EDH:@SECLEVEL=1";
```
~~to nginx config but it didn't work.~~ edit: not sure where I tried it (i think on centos 8) but it worked on 20.04

So if you can somehow enable TLS 1.0/1.1 then it should work.

Alternatively you could just use cloudflare, which supports it.

ledlamp:
this worked on ubuntu 20.04

closes #107 

todo maybe update the readme for 20.04 as well, cuz now u can just `apt install python3.9 python3-pip`. also python3.9-dev and python3.9-distutils don't seem to be needed

o btw instead of `git submodule init && git submodule update` just git clone `--recursive`.

o and gulag builds oppai for u so that part can be omitted too

### DIFF
--- a/ext/nginx.conf
+++ b/ext/nginx.conf
@@ -20,6 +20,7 @@ server {
 
 	ssl_certificate     /etc/letsencrypt/live/cmyui.xyz/fullchain.pem;
 	ssl_certificate_key /etc/letsencrypt/live/cmyui.xyz/privkey.pem;
+	ssl_ciphers "EECDH+AESGCM:EDH+AESGCM:AES256+EECDH:AES256+EDH:@SECLEVEL=1";
 
 	client_max_body_size 64m;
 
@@ -39,6 +40,7 @@ server {
 
 	ssl_certificate     /etc/letsencrypt/live/cmyui.xyz/fullchain.pem;
 	ssl_certificate_key /etc/letsencrypt/live/cmyui.xyz/privkey.pem;
+	ssl_ciphers "EECDH+AESGCM:EDH+AESGCM:AES256+EECDH:AES256+EDH:@SECLEVEL=1";
 
 	location / {
 		default_type image/png;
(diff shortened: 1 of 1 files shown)


cmyui:
> todo maybe update the readme for 20.04 as well, cuz now u can just apt install python3.9 python3-pip. also python3.9-dev and python3.9-distutils don't seem to be needed

Hmm, I agree it's kinda ugly to add a ppa when it's not required on 20.04 but this would cause a bigger issue on other distros where it's not available since it would actually be a roadblock for less experienced people.. Perhaps leaving the add-apt-repository line there but commented out with a little "you may have to use this if your distro doesn't have python3.9 available" comment would be okay.

> o btw instead of git submodule init && git submodule update just git clone --recursive.

Oh this is dope, didn't know about that ty.

cmyui:
As for the actual commit though, do you mind explaining why you picked these ciphers? I'm not too familiar with this cryptography stuff lol.

Does this force osu! to use a newer version, or force the server to use an older version? What do osu! & the server normally agree to use if this isn't enforced?

ledlamp:
I don't know anything about this cipher stuff, I just copied it from [here](https://askubuntu.com/a/1255186/516841) and found it to be the only solution that enabled TLSv1.0. We know that osu uses TLS 1.0 and nginx on ubuntu 20.04 only supports TLS 1.2 and 1.3. Adding this line enables TLS 1.0, 1.1, 1.2, and 1.3 on ubuntu 20.04. And by placing it inside the server blocks, only these vhosts are affected. It could be placed outside as well, but it would affect all vhosts.

I assume that having TLS 1.0 available means osu will be able to connect, but I have not tested it; we should test it to be sure before implementing this

ledlamp:
Tested and working with osu

PR #124, opened 2021-07-30
gulag systemd service

ledlamp:
recreated pr

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,8 +125,14 @@ nano config.py
 # have completely configured the file.   #
 ##########################################
 
-# start the server
+# test the server
 ./main.py
+
+
+# you can install the systemd service to run gulag as a daemon that starts at boot
+sudo cp ext/gulag.service /etc/systemd/system/gulag.service
+sudo nano /etc/systemd/system/gulag.service
+sudo systemctl daemon-reload && sudo systemctl enable --now gulag.service
 ```
 
 Directory Structure
--- a/ext/gulag.service
+++ b/ext/gulag.service
@@ -18,9 +18,8 @@ Group=gulag
 WorkingDirectory=/srv/gulag/
 ExecStart=/srv/gulag/main.py
 
-# uncomment if you want to output log to a separate file instead of the system journal
-#StandardOutput=file:/var/log/gulag.log
-#StandardError=file:/var/log/gulag.log
+StandardOutput=file:/var/log/gulag.log
+StandardError=file:/var/log/gulag.log
 
 
 [Install]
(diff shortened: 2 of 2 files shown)


cmyui:
Right so we've talked about this a bit in discord but lets make a thread of possible issues & improvements; basically I'm open to adding this functionality and even making it the recommended way to run gulag if it works better, but I'm a bit worried there may be some things in the codebase currently that weren't designed with this in mind that should be reworked a bit.

The immediate thing that comes to mind for me is logging to stdout;
- Do you think theres a better way to do logging in this case? (Should we switch logging completely to go to a specified file in config without colouring? I know tools like `tail -f` don't work with colour codes so I'm sure atleast some basic changes are needed to make this feel correctly supported)
- Is there any way to "bind" the terminal to a running service to see & interact with it as if you were just running it something like a tmux window? (since that's my normal usecase). If this is a thing, perhaps we can just expect users to use that and keep logging as-is?

ledlamp:
Systemd can pipe output to a file:
https://github.com/ledlamp/gulag/blob/patch-4/ext/gulag.service#L22-L23

Tail works fine with color...


The only issues with daemonizing are these two input prompts; this totally useless "press enter to continue" on first run:
https://github.com/cmyui/gulag/blob/419be0336bf46cad4a891bab3d6b53db552fac86/utils/updater.py#L49
And this "press enter to exit" on sql update failure:
https://github.com/cmyui/gulag/blob/419be0336bf46cad4a891bab3d6b53db552fac86/utils/updater.py#L173

I think both of these should be removed, they seem totally pointless. But you could make it check for the presency of a tty if you really wanted to.

Besides these, gulag doesn't take any input as far as I am aware, so there is no problem with daemonizing. 


I'm not sure on the best way of attaching a terminal to a daemon, but you can have the service screen or tmux if you want. This isn't usually done... daemons typically have an api and a command line tool for this sort of thing. Do you plan to add a command line interface to gulag?

ledlamp:
tldr remove the two input()'s and it's perfect

gamerboy80:
> tldr remove the two input()'s and it's perfect

or have an env variable or config option to not have user input things

ledlamp:
it can check if stdin is attached to a terminal to do that automatically. but right now the only input prompts are absolutely pointless.

cmyui:
Hmmmm I hadn't really considered the cli thing, but yea that's probably the right direction to go then.

If they want to actively look at the server they can just tail it or run directly then, but I'll leave the option to run it directly open since I'd imagine debugging is a lot simpler of a process.

I'll remove the inputs now though.

ledlamp:
If outputting to journal, colors are stripped:

cmyui:
will probably pull this but i just want to test it myself first just in case i want to change anything in it/gulag, and i'm quite busy irl atm >.<

PR #125, opened 2021-07-31
Delete nginx check

ledlamp:
nginx is NOT a dependency.

Docker users have to hack around it
https://github.com/ledlamp/gulag-dockerized/blob/1eb4531d66afb0d153f3d7482830d0e395ae18cb/source/gulag.Dockerfile#L6
https://github.com/catgirlmoe/gulag/blob/0c3defbc98d8bb2f7050ade3a6507bc6d4c96145/Dockerfile#L22-L23
y dont these embed >:(


Atakku:
Yeah I complained about this before

PR #126, opened 2021-07-31
refactor installation guide

ledlamp:
- supports 18.04 & 20.04, adds some important information.
- shows how to prepare mysql for gulag
- I changed the certbot command to use nginx plugin with list of domains instead of manual dns for wildcard; This way user only needs to run the one command and automatic renewal will work. The other way requires more complication and manual renewal.
- systemd service info

let me know what you think; may want revising

this pr depends on other pr's.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ the new stack; the code will need to be of comparable quality to what you see he
 
 # Installation Guide
 
-## Important notes:
+**Important notes:**
 - gulag currently does not work on most other distros besides ubuntu 18.04 and 20.04 due to an issue with the liboppai library.
 - osu uses the old deprecated TLSv1.0 so you will need to make sure your reverse proxy supports TLSv1.0. The included nginx configuration should enable TLSv1.0 on 20.04.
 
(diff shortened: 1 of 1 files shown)


tsunyoku:
just my 2 cents, while this is cool and all i dont think the plan is to make gulag so user friendly. as far as i know, cmyui's general goal with gulag is to make it dev friendly and so the user should have enough knowledge to need such an indepth guide. maybe drop some detail in this?

tsunyoku:
to not need*

ledlamp:
this is "dev" friendly. a lot of "dev"s don't know a lot of things. even cmyui himself didn't know some of this. I don't see what good it would be to reduce detail, we should tell them everything they need to know. it's not complicated, but as someone who never uses mysql for example, it's nice to have the readme tell how to init mysql instead of having to google it.

cmyui:
I've been wanting to do some changes to the readme as well, so I'll use your changes as a base since it's definitely better.

Will pull this then commit my changes together when I have some more time.

PR #133, opened 2021-09-10
Implemented peace-performance for non std modes.

alowave223:
Full implementation of peace-performance for taiko, catch and mania. Also updated /np and !with command to use peace-performance.


cmyui:
Thanks for the PR! :)

PR #139, opened 2021-10-20
It's a plain text, fstring isn't needed here

def750:
Credit to lenforiee, he noticed :trollface:


RealistikDash:
:trollface:

cmyui:
hahaha thanks for that

PR #144, opened 2021-11-20
Fix score submission

tsunyoku:
In commit https://github.com/cmyui/gulag/commit/b1694fc08a235ffd423b850987514b632cabb423, some cleanups in score submission caused the indexes of score data to be inaccurate.


tsunyoku:
After testing with some people in Discord, it seems these indexes don't work so it's probably better looking into exactly what changed before merging as some fixes are probably needed

cmyui (approved):

cmyui:
is fine, thanks :)

PR #157, opened 2022-01-04
Add CORS middleware, config for accept fetching data from frontend (guweb)

Varkaria:
for fixing `Access to XMLHttpRequest at 'https://domain/endpoint' from origin 'frontend_domain' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.`

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -24,6 +24,9 @@ DATADOG_APP_KEY=
 
 DEBUG=False
 
+# cors config for fetching api from frontend
+CORS_ORIGINS=http://localhost:3000,https://cmyui.xyz
+
 # redirect beatmaps, beatmapsets, and forum
 # pages of maps to the official osu! website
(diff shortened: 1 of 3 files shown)


tsunyoku:
we already have some unpushed commits relating to cors and other middlewares, but i’ll keep this open and unmerged for now just in case.

cmyui:
Oh nice! Thanks I'll merge this soon with the other PRs.

cmyui:
few things that came to mind,
1. are the `api.` and `osu.` subdomain origins required for CORS as well?
2. is :3000 for the frontend? i believe guweb is configured to run on 8000
3. we may be able to remove the config option by doing it dynamically in python with `app.settings.DOMAIN`

cmyui:
not sure about this one, don't think we should be hacking around things with cors, there may be a better fix on FE side here?

PR #161, opened 2022-01-04
Move settings outside of app

tsunyoku:
this allows us to make migration tools without needing to access app (doesn't work)
pr'd rather than merged as i know cmyui wants to review this first, just in case :)


cmyui:
there are other tools using manual DB_DSN still i believe

cmyui:
💃🏻 💃🏻 💃🏻 💃🏻 💃🏻 💃🏻 💃🏻

PR #162, opened 2022-01-05
Use pybind11 for oppai-ng bindings

### LINKED ISSUE #138: cmyui/oppai-ng wrapper causes a segmentation fault on most distros besides Ubuntu
So this was talked about before but it ought to be in the issue tracker. As i understand it, gulag was made to use the oppai library c bindings instead of executing the oppai binary as a child process, which created this issue.

Workarounds require running gulag in an ubuntu environment one way or another. Or debian 11 is the only other distro known to work.

tsunyoku:
Currently, we are using a ctypes wrapper around oppai-ng for speed, which is great, however prone to bugs (see https://github.com/cmyui/gulag/issues/138).

I made a new wrapper for oppai based on cmyui’s fork made with pybind11 to directly compile a python binding eliminating any possible bugs, and (partly) supporting Windows on the way.

I created this pull request instead of merging as I need some testers to confirm this branch is stable, so please reach out in the gulag Discord! Note when testing that oppai is only used for osu!standard in gulag.

closes #138

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -60,7 +60,7 @@ DATA_PATH = Path.cwd() / ".data"
 ACHIEVEMENTS_ASSETS_PATH = DATA_PATH / "assets/medals/client"
 DEFAULT_AVATAR_PATH = DATA_PATH / "avatars/default.jpg"
 DEBUG_HOOKS_PATH = Path.cwd() / "_testing/runtime.py"
-OPPAI_PATH = Path.cwd() / "oppai-ng"
+OPPAI_PATH = Path.cwd() / "oppai_ng"
 
 
 useful_keys = (Keys.M1, Keys.M2, Keys.K1, Keys.K2)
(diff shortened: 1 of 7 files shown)


app/utils.py
```diff
@@ -430,38 +430,35 @@ def ensure_directory_structure() -> int:
 
 def ensure_dependencies_and_requirements() -> int:
     """Make sure all of gulag's dependencies are ready."""
-    if not OPPAI_PATH.exists():
+    if not OPPAI_PATH.exists() or not any(OPPAI_PATH.iterdir()):
```
cmyui, line 433: kinda weird should mention pybind specifically

app/utils.py
```diff
@@ -430,38 +430,35 @@ def ensure_directory_structure() -> int:
...
-            args=["git", "submodule", "update"],
-            stdout=subprocess.DEVNULL,
-            stderr=subprocess.DEVNULL,
-        )
-        if exit_code := p.wait():
-            log("Failed to update git submodules.", Ansi.LRED)
-            return exit_code
-
-    if not (OPPAI_PATH / "liboppai.so").exists():
+    if not any(Path(OPPAI_PATH).glob("*.so")):  # TODO: could not be .so if windows?
```
cmyui, line 444: kinda weird

PR #164, opened 2022-01-29
Devserver is supported

def750:
This can lead to misunderstanding. Devserver is in stable since like 8 months.


cmyui:
thanks, didn't notice this :)

PR #170, opened 2022-02-13
display clan id in api request

tsuuki:
idk y this wasn't done before

### DIFF
--- a/app/api/domains/api.py
+++ b/app/api/domains/api.py
@@ -145,14 +145,14 @@ async def api_get_player_info(
     if username:
         user_info = await app.state.services.database.fetch_one(
             "SELECT id, name, safe_name, "
-            "priv, country, silence_end "
+            "priv, clan_id, country, silence_end "
             "FROM users WHERE safe_name = :username",
             {"username": username.lower()},
         )
     else:  # if user_id
         user_info = await app.state.services.database.fetch_one(
             "SELECT id, name, safe_name, "
-            "priv, country, silence_end "
+            "priv, clan_id, country, silence_end "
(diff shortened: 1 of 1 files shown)


cmyui:
good catch, but it seems a bit weird to only include clan id. i think if we're going to include clan info, we should include the full spread of what's useful for frontend.

probably should be done by fetching the clan from the player, and calling format_clan_basic() on it.

if you're up for it, try it out; if not, I can add to this PR later and we can merge it :)

tsuuki:
i'm a tad busy but i'll try it out if you don't in the next coming days

app/api/domains/api.py
```diff
@@ -145,14 +145,14 @@ async def api_get_player_info(
     if username:
         user_info = await app.state.services.database.fetch_one(
             "SELECT id, name, safe_name, "
-            "priv, country, silence_end "
+            "priv, clan_id, country, silence_end "
```
alowave223, line 148: for what is even silence_end put here?

cmyui: on osu/ripple frontend u can see how long someone is silenced for, so i thought i'd include it, public enough data

alowave223:
i think, it would be much better if we use `Player` object instead of raw sql data, we have `from_cache_or_sql` function and in my opinon it looks better and scalable in future without any need at editing sql and parsing raw sql data, also change a `format_player_basic` little, and add new properties `safe_name` and `silence_end` since we need to get user_info either way, also we already have clan parsed with `format_clan_basic` in it.

cmyui:
going to merge this and open an issue to add the clan object

PR #180, opened 2022-03-15
rename server implementation to bancho.py

cmyui:


migrations/base.sql
```diff
@@ -149,7 +149,7 @@ create table maps
 
 create table mapsets
 (
-	server enum('osu!', 'gulag') default 'osu!' not null,
+	server enum('osu!', 'gulag') default 'osu!' not null, # TODO: change name
```
cmyui, line 152: will need a migration for these

PR #191, opened 2022-03-19
Add support for more common beatmap mirror APIs

zvyap:
Nerinyan's API is a lot better then other beatmap API in my experience

 - [x] Support Newest, Top+Rated, Most+Played [Only on Nerinyan]

 - [x] Tested Chimu
 - [x] Tested Cheesegull
 - [x] Tested Nerina


.env.example
```diff
@@ -10,6 +10,12 @@ REDIS_DSN=redis://localhost
 OSU_API_KEY=
 
 DOMAIN=cmyui.xyz
+
+# Here is how mirror type work [also official mirror url]
+# 1 = Chimu api, https://api.chimu.moe/v1
+# 1 = Cheesegull api, https://storage.ripple.moe/
+# 3 = Nerina api, https://api.nerina.pw/
+MIRROR_TYPE=1
```
cmyui, line 18: i wonder if it's possible to use `MIRROR_TYPE(str, Enum)` so we can have this as a string rather than an int?

app/api/domains/osu.py
```diff
@@ -1699,13 +1766,13 @@ async def get_osz(
...
-        query_str = f"download/{map_set_id}?n={int(not no_video)}"
+    if not MIRROR_TYPE.CHEESEGULL:
+        query_str = f"{CURRENT_MIRROR_TYPE.download_path}/{map_set_id}?{CURRENT_MIRROR_TYPE.no_video_param}={int(not no_video)}"
     else:
-        query_str = f"d/{map_set_id}"
+        query_str = f"{CURRENT_MIRROR_TYPE.download_path}/{map_set_id}"
 
     return RedirectResponse(
-        url=f"{app.settings.MIRROR_URL}/{query_str}",
+        url=f"{app.settings.MIRROR_URL}{query_str}",
```
cmyui, line 1775:
```suggestion
        url=f"{app.settings.MIRROR_URL}/{query_str}",
```

app/api/domains/osu.py
```diff
@@ -431,28 +474,37 @@ async def osuSearchHandler(
     mode: int = Query(..., alias="m", ge=-1, le=3),  # -1 for all
     page_num: int = Query(..., alias="p"),
 ):
-    if USING_CHIMU:
-        search_url = f"{app.settings.MIRROR_URL}/search"
-    else:
-        search_url = f"{app.settings.MIRROR_URL}/api/search"
+    search_url = f"{app.settings.MIRROR_URL}{CURRENT_MIRROR_TYPE.search_path}"
```
cmyui, line 477:
```suggestion
    search_url = f"{app.settings.MIRROR_URL}/{CURRENT_MIRROR_TYPE.search_path}"
```

app/api/domains/osu.py
```diff
@@ -405,23 +406,65 @@ async def lastFM(
...
+    CHEESEGULL = 2
+    NERINA = 3
+
+    @functools.cached_property
+    def search_path(self) -> str:  # more cleaner way to code these
+        if self == self.CHIMU:
+            return "/search"
+        if self == self.CHEESEGULL:
+            return "/api/search"
+        return "/search"  # NERINA
```
cmyui, line 423:
```suggestion
    def search_path(self) -> str:  # more cleaner way to code these
        if self == self.CHIMU:
            return "search"
        if self == self.CHEESEGULL:
            return "api/search"
        return "search"  # NERINA
```

app/api/domains/osu.py
```diff
@@ -405,23 +406,65 @@ async def lastFM(
...
+            return "SetID"
+        return "id"  # NERINA
+
+    @functools.cached_property
+    def download_path(self) -> str:
+        if self == self.CHIMU:
+            return "/download"
+        if self == self.CHEESEGULL:
+            return "/d"
+        return "/d"  # NERINA
```
cmyui, line 439:
```suggestion
    @functools.cached_property
    def download_path(self) -> str:
        if self == self.CHIMU:
            return "download"
        if self == self.CHEESEGULL:
            return "d"
        return "d"  # NERINA
```

cmyui:
thanks for the work though! this looks good :)

.env.example
zvyap: > i wonder if it's possible to use `MIRROR_TYPE(str, Enum)` so we can have this as a string rather than an int?

I think it work? I use number because it easy for user settings

I can change it to string if your want

cmyui: yeah i'd prefer it as a string just for readability - thanks!

aeongdesu:
Can you change nerina mirror server name to nerinyan? it is official name

zvyap:
> Can you change nerina mirror server name to nerinyan? it is official name

I know but the mirror name call Nerina right?
I follow the mirror name, I guess I will just ask the Nerinyan staff about it

zeee2:
Please change Neriyan's api address api.nerina.pw to api.nerinyan.moe

zvyap:
> Please change Neriyan's api address api.nerina.pw to api.nerinyan.moe

Ya I know, will change later

PR #200, opened 2022-03-21
Add leaderboard caching

### LINKED ISSUE #203: Add leaderboard caching
Leaderboards are one of the primary resources requested by an osu! client.

We plan to make significant improvements to the efficiency of the server by caching the all types of in-game leaderboards, like has already been done for other entities such as beatmaps.

tsunyoku:
This PR aims to add efficient caching for leaderboards to bancho.py.

The cache will get ALL scores in sql for a given beatmap, and is sorted/validated as per use case. This allows to easily remove/add scores for any reason (restrictions/unrestrictions, score submissions) without the need for extra communication or sql calls, further boosting the caching's efficiency.

This PR is untested as I wrote this while away from home, so I will ask for a review and test when I'm able to. Until it has been tested, I will not merge.

closes #203


app/objects/leaderboard.py
```diff
@@ -0,0 +1,97 @@
...
+
+        leaderboard = Leaderboard(mode)
+
+        rows = await app.state.services.database.fetch_all(
+            "SELECT id, map_md5, userid, pp, score, "
+            "max_combo, mods, acc, n300, n100, n50, "
+            "nmiss, ngeki, nkatu, grade, perfect, "
+            "status, mode, play_time, "
+            "time_elapsed, client_flags, online_checksum "
+            "FROM scores WHERE map_md5 = :map_md5 AND status = 2"
```
cmyui, line 40:
```suggestion
            "FROM scores WHERE map_md5 = :map_md5 AND status = 2 "
```

app/objects/leaderboard.py
```diff
@@ -0,0 +1,97 @@
...
+        return 0
+
+    async def remove_user(self, user_id: int) -> None:
+        result = await self.find_user_score(user_id)
+
+        if result is not None:
+            self.remove_score_index(result["rank"] - 1)
+
+    async def sort(self) -> None:
+        async with self.lock:
```
cmyui, line 83: perhaps this lock should be used at a higher level and not within the sort function?

would be less side-effectey, would make it more explicit when using it, doesn't need to be async

app/objects/leaderboard.py
```diff
@@ -0,0 +1,97 @@
...
+        await leaderboard.sort()
+        return leaderboard
+
+    async def remove_score_index(self, index: int) -> None:
+        async with self.lock:
+            self.scores.pop(index)
+
+    async def find_user_score(
+        self,
+        user_id: int,
```
cmyui, line 58:
```suggestion
        user_id: int
```

app/objects/leaderboard.py
```diff
@@ -0,0 +1,97 @@
...
+    ) -> Optional[UserScore]:
```
cmyui, line 59:
```suggestion
    async def find_user_score(self, user_id: int) -> Optional[UserScore]:
```

cmyui, line 59:
```suggestion
    async def find_user_score(self, user_id: int) -> Optional[UserScore]:
```

cmyui (commented):
nice work. here's some stuff i noticed from github - will dive into this in vscode shortly

app/objects/leaderboard.py
tsunyoku: black 👍🏻

cmyui: if u add the `,`, it will do that - this is probably the only control you have with black

cmyui:
needs testing, will reach out to a few people using the sevrer in production.

on our side, we should get some >several thousand score leaderboards from akatsuki for testing purposes - would be super useful for cleaning up, optimizing, and testing here.

0i8:
why was this closed?

tsunyoku:
after running similar caching on akatsuki for over a year, this code is quite bad but also caching leaderboards is not really worth it... akatsuki is the largest private server, so if any server were to benefit from cache it would be us and yet we haven't. all it did was decrease load on the database and increase our memory usage tenfold, but the reasons for leaderboards being bad without caching in the first place was ripple's unoptimized tables and indexes. if your database is designed well, the risks of caching (there are plenty) are far worse than the reward you get for caching leaderboards

basically - don't cache your leaderboards!

7ez:
> after running similar caching on akatsuki for over a year, this code is quite bad but also caching leaderboards is not really worth it... akatsuki is the largest private server, so if any server were to benefit from cache it would be us and yet we haven't. all it did was decrease load on the database and increase our memory usage tenfold, but the reasons for leaderboards being bad without caching in the first place was ripple's unoptimized tables and indexes. if your database is designed well, the risks of caching (there are plenty) are far worse than the reward you get for caching leaderboards
> 
> basically - don't cache your leaderboards!

the RealistikOsu/USSR incident

0i8:
> after running similar caching on akatsuki for over a year, this code is quite bad but also caching leaderboards is not really worth it... akatsuki is the largest private server, so if any server were to benefit from cache it would be us and yet we haven't. all it did was decrease load on the database and increase our memory usage tenfold, but the reasons for leaderboards being bad without caching in the first place was ripple's unoptimized tables and indexes. if your database is designed well, the risks of caching (there are plenty) are far worse than the reward you get for caching leaderboards
> 
> basically - don't cache your leaderboards!

thank you for the insight

PR #215, opened 2022-04-08
Fixed check whether the user exists in changename

minisbett:
It used the name column instead of safe_name, resulting in an SQL error that the safe_name column is unique.

Example: User "test 123" exists and you want to change your name to "test_123"


7ez (changes requested):
remove the space between all checks passed comment and execute statement

cmyui (approved):

cmyui:
thanks!

PR #218, opened 2022-04-17
Fix OpenAPI docs not being generated

tsunyoku:
This was an internal issue where `.host()` would only interface Starlette, creating a `Host` object which FastAPI cannot deal with.

The fix this PR provides is a subclass of FastAPI which modifies the `.openapi()` call to include the routes from the `Host` objects it contains.

### DIFF
--- a/app/api/init_api.py
+++ b/app/api/init_api.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 import asyncio
 import os
 import pprint
+from typing import Any
 
 import aiohttp
 import orjson
@@ -15,7 +16,9 @@ from fastapi.requests import Request
 from fastapi.responses import ORJSONResponse
 from fastapi.responses import Response
 from starlette.middleware.base import RequestResponseEndpoint
+from fastapi.openapi.utils import get_openapi
 
+import starlette.routing
 import app.bg_loops
 import app.settings
 import app.state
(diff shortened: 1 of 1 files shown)


app/api/init_api.py
```diff
@@ -27,7 +30,38 @@
...
+    def openapi(self) -> dict[str, Any]:
+        if not self.openapi_schema:
+            routes = self.routes
+            starlette_hosts = [
+                host
+                for host in super().routes
+                if isinstance(host, starlette.routing.Host)
+            ]
+
+            for host in starlette_hosts:
```
cmyui, line 43:
```suggestion
            # XXX:HACK fastapi will not show documentation for routes
            # added through use sub applications using the Host class
            # (e.g. app.host('other.domain', app2))
```

cmyui:
one issue with this approach i've found - when using the documentation to execute queries, the api call will go to the url of the page that's currently open.

e.g., if you're on `c.ppy.sh` and test a request to `osu.ppy.sh/abc`, the request will go to `c.ppy.sh/abc`.

tsunyoku:
> one issue with this approach i've found - when using the documentation to execute queries, the api call will go to the url of the page that's currently open.
> 
> e.g., if you're on `c.ppy.sh` and test a request to `osu.ppy.sh/abc`, the request will go to `c.ppy.sh/abc`.
> 
> this is still likely something we can solve?

interesting. as for fixing, i don’t know if it’s possible: fastapi doesn’t care for what domain  a route is on, so we can’t even specify to fastapi which to use. we may have to resort to using include_router with a prefix for each service, and splitting the domains in nginx

minisbett:
Even if the query testing doesn't work having support for OpenAPI would still be really nice

tsunyoku:
fastapi does it by default but because we want to limit the domains we access it doesn't, my pr tried to fix it but it still ain't fully compatible with fastapi for some irritating reason

minisbett:
if I understood correctly, with your fix bpy still works, OAI works but the test feature doesn't? If so I mean, it's at least something, better than nothing

cmyui:
yeah, will merge it for now - perhaps we can do better in the future

PR #219, opened 2022-04-18
Typo in beatmap.py

def750:
Maybe I'm bad at english and if I'm wrong it's quite funny issue, but shouldn't this be "We keep"


minisbett:
this is the most pull request I've ever seen

minisbett (approved):
I approve

PR #223, opened 2022-04-27
Add Troubleshooting section

Mxnuuel:
so we don't get massive pinged for the same issue over and over again

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,11 @@ and you should see something along the lines of:
 
 ![ada](https://i.cmyui.xyz/ld-iZXysVXqwhM8.png)
 
+## Troubleshooting
+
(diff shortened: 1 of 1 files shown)


Mxnuuel:
I get that and that's just a "most case" scenario being reported when the server is not reachable.

Also, literally every other image is also hosted on a different server, i really don't care that much since i make sure everything just runs and also has been running for over 3 years now lmao.

alowave223:
**NEVER** had problems with flex encryption in 6 banco.py installations for different people, lol.

cmyui:
changed how ssl works in refactor branch - removing ssl from the server-side and assuming people use cloudflare flex by default, so this will no longer be accurate

PR #225, opened 2022-04-28
Fixed error on !user when specifying an offline user

minisbett:


app/commands.py
```diff
@@ -898,14 +898,20 @@ async def user(ctx: Context) -> Optional[str]:
     else:
         last_np = None
 
+    osu_version = (
+        p.client_details.osu_version.date
+        if hasattr(p.client_details, "osu_version")
```
7ez, line 903: do `if p.client_details is not None`

7ez (changes requested):
yes

app/commands.py
cmyui: `if p.online` should also suffice

cmyui:
```suggestion
        if p.online
```

app/commands.py
```diff
@@ -898,14 +898,20 @@ async def user(ctx: Context) -> Optional[str]:
...
+        else "Unknown"
+    )
     return "\n".join(
         (
             f'[{"Bot" if p.bot_client else "Player"}] {p.full_name} ({p.id})',
             f"Privileges: {priv_list}",
+            f"Donator status: {donator_status}",
```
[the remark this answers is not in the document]
cmyui: will just remove for now so we can merge. feel free to make another pr
```suggestion
```

app/commands.py
```diff
@@ -898,14 +898,15 @@ async def user(ctx: Context) -> Optional[str]:
     else:
         last_np = None
 
+    osu_version = p.client_details.osu_version.date if p.online else "Unknown"
```
cmyui, line 901:
```suggestion
    osu_version = p.client_details.osu_version.date if p.online else "Unknown"

```

cmyui (approved):

cmyui:
thank ya!

app/commands.py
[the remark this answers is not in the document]
minisbett: oh yeah lmao that's from my code I just copy pasted

PR #226, opened 2022-04-28
Give donator command, donator status in Player object etc.

7ez:


7ez:
all tested btw 👍

7ez:
I'll reopen the pull request later, as I want to use a more reliant parsing (pytimeparse) instead of regex parsing

PR #228, opened 2022-04-29
Make users unable to assign donator privileges using !addpriv

minisbett:
addpriv does not set a donator end therefore the background task will remove the donator privileges after some time anyway.
To hint users the proper way to assign donator privileges just make them unable to assign the privileges and return an error instead.

### DIFF
--- a/app/commands.py
+++ b/app/commands.py
@@ -1385,7 +1385,7 @@ async def addpriv(ctx: Context) -> Optional[str]:
 
     if not (t := await app.state.sessions.players.from_cache_or_sql(name=ctx.args[0])):
         return "Could not find user."
-    
+
     if bits & Privileges.DONATOR:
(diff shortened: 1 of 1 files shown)


app/commands.py
```diff
@@ -1407,6 +1407,9 @@ async def rmpriv(ctx: Context) -> Optional[str]:
     if not (t := await app.state.sessions.players.from_cache_or_sql(name=ctx.args[0])):
         return "Could not find user."
 
+    if bits & Privileges.DONATOR:
+        return "Please use the !givedonator command to assign donator privileges to players."
```
cmyui, line 1411: wrong function, this is rmpriv hehehe

if you fix i can merge

minisbett: Lmfao yeah gonna fix in like 9 hours

cmyui (approved):

PR #229, opened 2022-04-29
Add donor_end to player object, use pytimeparse consistently

7ez:
In the current state, the command won't run as there is no `donor_end` in the Player object.

### DIFF
--- a/app/commands.py
+++ b/app/commands.py
@@ -822,6 +822,7 @@ SHORTHAND_REASONS = {
     "au": "using 3rd party programs (auto play)",
 }
 
+
 @command(Privileges.MODERATOR, hidden=True)
 async def silence(ctx: Context) -> Optional[str]:
     """Silence a specified player with a specified duration & reason."""
@@ -1315,6 +1316,7 @@ async def debug(ctx: Context) -> Optional[str]:
     app.settings.DEBUG = not app.settings.DEBUG
     return f"Toggled {'on' if app.settings.DEBUG else 'off'}."
 
+
 # NOTE: these commands will likely be removed
 #       with the addition of a good frontend.
 str_priv_dict = {
@@ -1373,6 +1375,7 @@ async def rmpriv(ctx: Context) -> Optional[str]:
     await t.remove_privs(bits)
     return f"Updated {t}'s privileges."
 
+
 @command(Privileges.DEVELOPER, hidden=True)
 async def givedonator(ctx: Context) -> Optional[str]:
     """Gives donator to a specified player (by name) for a specified amount of time, such as '3h5m'."""
@@ -1396,6 +1399,7 @@ async def givedonator(ctx: Context) -> Optional[str]:
     )
 
     return f"Added {''.join(ctx.args[1:])} of donator status to {t}."
+
 
 @command(Privileges.DEVELOPER)
 async def wipemap(ctx: Context) -> Optional[str]:
--- a/app/commands.py
+++ b/app/commands.py
@@ -1408,7 +1408,7 @@ async def givedonator(ctx: Context) -> Optional[str]:
         {"end": timespan, "user_id": t.id},
     )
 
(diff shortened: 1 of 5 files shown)


7ez:
Code was also optimised a bit if you can even call it that

app/commands.py
```diff
@@ -1410,6 +1373,29 @@ async def rmpriv(ctx: Context) -> Optional[str]:
     await t.remove_privs(bits)
     return f"Updated {t}'s privileges."
 
+@command(Privileges.DEVELOPER, hidden=True)
```
minisbett, line 1379: I think admins should be able to assign donator privileges, I don't see a benefit in restricting the access to developers.

minisbett (changes requested):

app/commands.py
```diff
@@ -905,6 +900,7 @@ async def user(ctx: Context) -> Optional[str]:
         (
             f'[{"Bot" if p.bot_client else "Player"}] {p.full_name} ({p.id})',
             f"Privileges: {priv_list}",
+            f"Donator ends: {timeago.format(p.donor_end)}",
```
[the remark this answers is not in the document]
7ez: alr

app/commands.py
```diff
@@ -1411,6 +1376,31 @@ async def rmpriv(ctx: Context) -> Optional[str]:
...
+        timespan += int(time.time())
+    else:
+        timespan += t.donor_end
+
+    await app.state.services.database.execute(
+        "UPDATE users SET donor_end = :end WHERE id = :user_id",
+        {"end": timespan, "user_id": t.id},
+    )
+
+    return f"Added {''.join(ctx.args[1:])} of donator status to {t}."
```
cmyui, line 1401: should be just ctx.args[1]

app/commands.py
```diff
@@ -1411,6 +1376,31 @@ async def rmpriv(ctx: Context) -> Optional[str]:
...
+@command(Privileges.DEVELOPER, hidden=True)
+async def givedonator(ctx: Context) -> Optional[str]:
+    """Gives donator to a specified player (by name) for a specified amount of time, such as '3h5m'."""
+    if len(ctx.args) < 2:
+        return "Invalid syntax: !givedonator <name> <duration>"
+
+    if not (t := await app.state.sessions.players.from_cache_or_sql(name=ctx.args[0])):
+        return "Could not find user."
+
+    if not (timespan := timeparse("".join(ctx.args[1:]))):
```
cmyui, line 1388: should be just ctx.args[1] - `len(ctx.args) < 2` is asserted above

app/commands.py
7ez: i mean people may use spaces so incase they do it'll still support it

app/commands.py
7ez: that only checks if theres less than 2 args lol

cmyui: Oh right, didn't realize pytimeparse supported that
```py
>>> pytimeparse.parse(' 3h 5m')
11100
```

in this case, just make sure we're being consistent across all commands doing time parsing with this - if we allow spaces in one, we should allow spaces in all

7ez: i'd have to move the delay at the end on the shutdown command then

cmyui: ah sorry shutdown is okay as-is then, just worry about user-facing commands, dev commands are lower prio

7ez: what about silence though

7ez: do i just move duration to the end?

7ez: ill just leave it for now

app/commands.py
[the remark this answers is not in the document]
minisbett: I wanted to do that once this pr is merged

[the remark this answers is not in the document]
7ez: its already been done though

7ez:
@cmyui all good now 👍

cmyui (approved):

PR #234, opened 2022-05-02
optimization for donator info in !user command

minisbett:


cmyui:
can u pls just name things what they actually are, it just ruins the log of commits/requests - treat git commit names, comments, pr names as logs

cmyui:
i won't really like squash merging commits but i do it in these cases to keep the log clean
image

cmyui:
i'm glad you're helping out - i don't mean to set the wrong tone but i try my best to keep it clean :^)

minisbett:
sure sure I just got kinda lazy because it was such a small change

PR #240, opened 2022-06-02
Do not update ranks for restricted players

alowave223:


minisbett:
What difference does this make?

tsunyoku:
> What difference does this make?

if they're restricted and submit a score, they're going to be re-added to leaderboards without this

minisbett:
I think I already tested that and I'm pretty sure the restricted player did not manipulate the rankings of players in any way after the restriction

PR #245, opened 2022-06-19
globally fix DTNC bug

juup1ter:
YEP


cmyui:
hey - if you need to filter invalid mod combinations, you should call the Mods.filter_invalid_combos() method

juup1ter:
shouldn't we filter invalid mod combinations already, rather than calling a method?

minisbett:
well, the expected output of the method is to convert the mod enum into a list of it's readable short names. And since NC always has the DT flag set, it's DTNC

PR #205, opened 2022-03-24
v5.0.0 - Overall project restructuring & refactoring

cmyui:
# v5.0.0 - Overall project restructuring & refactoring
This one comes with quite a few updates and improvements to the codebase.
(implementing more of my learnings from time away)

## decoupling of logic & data into separate (functional) api layers
One of the main issues facing the codebase is it's tight coupling of logic & data.

Let's take the Beatmap class as an example - it;
- defines [the representation of an osu! beatmap](https://github.com/osuAkatsuki/bancho.py/blob/master/app/objects/beatmap.py#L272-L362)
- provides [methods to interact with specific instances](https://github.com/osuAkatsuki/bancho.py/blob/master/app/objects/beatmap.py#L375-L558)
- contains [the logical functionality for managing osu! beatmaps](https://github.com/osuAkatsuki/bancho.py/blob/master/app/objects/beatmap.py#L616-L768)

If you look through bancho.py's codebase, you'll see this recurring theme implemented in many inconsistent ways.
(see if you can find the similarities & differences in [the Player class](https://github.com/osuAkatsuki/bancho.py/blob/master/app/objects/player.py#L195-L1169) for example)

This PR aims to decouple structural representation, data providers, and business logic into 3 separate api levels.

### data structures will be represented as "models"
Models represent the format of a resource.
They allow you to think only about the structure of a resource - the attributes within it and its properties.

An example of a model may be the representation of a beamtap - for example;
```py
from dataclasses import dataclass

@dataclass
class Beatmap:
    id: int
    set_id: int
    md5: str
    ranked_status: int

    circle_size: float
    health_drain: float
    approach_rate: float
    overall_difficulty: float
```
(you may also define simple properties, but nothing involving complex logic or external data points)

### data providers will be represented as "repositories"
Repositories are an abstraction around the interaction with a resource.
They allow you to think of the interactions with the data providing service, as opposed to the actual implementation details of the underlying services - proving a standardized API for your business-logic layers.

Continuing with the theme of beatmaps, a simple repository may look something like
```py
from typing import Mapping
from typing import Optional

import models
import services

## in-memory cache

id_cache: Mapping[int, models.Beatmap] = {}
md5_cache: Mapping[str, models.Beatmap] = {}

## reads

async def fetch_by_id(id: int) -> Optional[models.Beatmap]:
    if beatmap := id_cache.get(id):
        return beatmap

    beatmap_row = await services.database.fetch_one(
        "SELECT * FROM beatmaps WHERE id = :id",
[...]

PR #246, opened 2022-06-28
Change avatar serving to be done through NGINX

Airiuwu:
Why go through nginx to get to code to serve an avatar when you can just go to nginx to serve it?


7ez:
because gif support

Airiuwu:
gifs cant work in game so no point

7ez:
they can on website

minisbett:
And this way you get much better control over serving avatars, as well as providing much more possibility for customization or adjustment. Yes, not many people will change something on bpy but why would you change it when it has no disadvantage, just that tiny advantage? Seems like pointless effort to me without any effect on the outcome.
Another disadvantage is the inconvenience of using bpy with other webservers besides nginx since then you'd have to find out how a config file for that would look like, if the webserver even can provide that functionality (which most big webservers like apache do tho).

Personally I'd say that this is something that should be included in the bpy code such that it can be seen as a whole independent from any third party software. But this opinion on how to structure a project and it's infrastructure of functionality varies from developer to developer.

cmyui:
hey i'll be doing this in my refactor branch, #205 along with some other improvements to nginx - i'll reach out shortly if you'd like to move this PR over to that branch

cmyui:
changes look good though

cmyui:
open something like this (tested) to #205 and i'll merge it

PR #249, opened 2022-07-05
Fix console image in Chinese translation

hedgehog-qd:
Fix console image in Chinese translation


cmyui:
ah nice hahaha i already forgot

cmyui (approved):

PR #251, opened 2022-07-08
If no osu!api key is defined, use Kitsu's API to get beatmap metadata

HorizonCode:
If the user has no osu!api key defined, then use the Kitsu API.


alowave223:
You can get and use your api key even if you has been restricted, there's no reason to relay on 3rd party service

7ez:
> You can get and use your api key even if you has been restricted, there's no reason to relay on 3rd party service

also even if your account is disabled

alowave223 (commented):
Overall, relaying on 3rd party service can't be same 1:1 to regular osu!api and can lead to unexpected problems, outage or maintenance of the service which can negatively reflect on user expirience.

HorizonCode:
> Overall, relaying on 3rd party service can't be same 1:1 to regular osu!api and can lead to unexpected problems, outage or maintenance of the service which can negatively reflect on user expirience.

I bet you didn't even try it, and yet you are judging about it.
I have this running with USSR on my personal Server since about 1 Month and i don't experience any issues.

alowave223:
> > Overall, relaying on 3rd party service can't be same 1:1 to regular osu!api and can lead to unexpected problems, outage or maintenance of the service which can negatively reflect on user expirience.
> 
> I bet you didn't even try it, and yet you are judging about it. I have this running with USSR on my personal Server since about 1 Month and i don't experience any issues.

Yes, you're right i did not try it but i just spitting my opinion on using 3rd party services at all, however in this situation using of 3rd party services it's absolutelly unnecessary

HorizonCode:
> > > Overall, relaying on 3rd party service can't be same 1:1 to regular osu!api and can lead to unexpected problems, outage or maintenance of the service which can negatively reflect on user expirience.
> > 
> > 
> > I bet you didn't even try it, and yet you are judging about it. I have this running with USSR on my personal Server since about 1 Month and i don't experience any issues.
> 
> Yes, you're right i did not try it but i just spitting my opinion on using 3rd party services at all, however in this situation using of 3rd party services is absolutely unnecessary

But though this is an option for users who don't want to enter their osu!api key, or may experience issues with the main api.
If you don't want to use the Kitsu api then just enter the api key and it's all good.
This is just an alternative.

alowave223:
> > > > Overall, relaying on 3rd party service can't be same 1:1 to regular osu!api and can lead to unexpected problems, outage or maintenance of the service which can negatively reflect on user expirience.
> > > 
> > > 
> > > I bet you didn't even try it, and yet you are judging about it. I have this running with USSR on my personal Server since about 1 Month and i don't experience any issues.
> > 
> > 
> > Yes, you're right i did not try it but i just spitting my opinion on using 3rd party services at all, however in this situation using of 3rd party services is absolutely unnecessary
> 
> But though this is an option for users who don't want to enter their osu!api key, or may experience issues with the main api. If you don't want to use the Kitsu api then just enter the api key and it's all good. This is just an alternative.

If you had written this originally in the description of the pr, there would be no questions to you because the concept of "if not defined" is quite vague. But anyway, i'll stay stick to my opinion on that because i can't think of real world reason why you would not use regular osu api.

tsunyoku:
I understand where alo is coming from, the plan with bpy really is to have as little reliance as possible on any external sources - if someone doesn’t provide an API key and kitsu is offline or shut down or something else which prevents it from working, what do we do now? bancho will be running for aslong as the game does, which is aslong as we need to guarantee that we have a working API. that is why it is better to force an API key and use bancho, it is a safety guarantee.

cmyui:
hm tbh i think it's a valid usecase for people that aren't comfortable hooking their osu api key up to a server where any user can abuse it, but i might add a warning or something at startup to people are aware they'll have slightly less-reliable beatmap info

cmyui:
also, perhaps i can add option to hook up multiple osu api keys and ratelimits on a per-key basis to help with abuse potential in general

cmyui:
> if someone doesn’t provide an API key and kitsu is offline or shut down or something else which prevents it from working, what do we do now?

they provide an osu!api key and have the same functionality as now - this is only an additive layer for if one is not provided

tsunyoku:
> > if someone doesn’t provide an API key and kitsu is offline or shut down or something else which prevents it from working, what do we do now?
> 
> they provide an osu!api key and have the same functionality as now - this is only an additive layer for if one is not provided

my hypothetical was if they didn't provide one either though, we should still force an api key either way if the user wants any reliable beatmap info

cmyui (approved):
+1 👍🏻

PR #255, opened 2022-07-18
redis

memsdm05:
redis


cmyui (changes requested):
hey, looks good - could you move it  directly below mysql though? seems more appropriate ^^

cmyui (approved):

PR #257, opened 2022-07-22
Add db column to track when a user added map as a favourite; add redirect for beatmapset discussion page

hwk929:
Favourite beatmaps on bancho will show in the order that you favourite them:
image

I'd like to do the same on bancho.py, but it isn't possible as the only two colums are `userid` and `setid`. Instead, the order is based on when the beatmap itself was inserted into the `maps` database. My only other option is to order by the beatmap's `last_update`. Adding `fav_time` allows me to sort by the time the map was favourited, and gives more flexibility for future queries. I think it would be a pretty useful addition for both me, and any other frontend/api projects.

I also added a redirect for beatmap threads. I'm aware the code will likely be [removed](https://github.com/osuAkatsuki/bancho.py/blob/592d386f70aee2c1c0833250642ea0b5ea97cd52/app/api/domains/osu.py#L1658) but I can't see a frontend making use of this anytime soon.

```sql
ALTER TABLE favourites ADD fav_time INT DEFAULT 0 NOT NULL;
```

### DIFF
--- a/app/api/domains/osu.py
+++ b/app/api/domains/osu.py
@@ -1665,6 +1665,7 @@ if app.settings.REDIRECT_OSU_URLS:
     for pattern in (
         "/beatmapsets/{_}",
         "/beatmaps/{_}",
+        "/beatmapsets/{_}/discussion",
         "/community/forums/topics/{_}",
     ):
         router.get(pattern)(osu_redirect)
--- a/app/api/domains/osu.py
+++ b/app/api/domains/osu.py
@@ -327,7 +327,7 @@ async def osuAddFavourite(
 
     # add favourite
     await app.state.services.database.execute(
-        "INSERT INTO favourites VALUES (:user_id, :set_id)",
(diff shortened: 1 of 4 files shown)


cmyui:
Hey - thanks for the PR, I agree this is a good idea!

Two requests;
1. I'd like if the new column could be named `created_at` - in most tables in the future I'll be using `created_at` and `updated_at` for fields like this - I'd like it to be consistent across all tables.
2. You've added this to the base database structure, but you'll need to [bump the project version](https://github.com/osuAkatsuki/bancho.py/blob/master/app/settings.py#L58) and [add a new database migration](https://github.com/osuAkatsuki/bancho.py/blob/master/migrations/migrations.sql) for the new version for this update to be applied to people who have already installed the server.

cmyui:
I'll also likely be moving all my unix timestamp columns to datetime in the near future - feel free to do so if you'd like, or I can do them all at once.

cmyui (commented):
thanks for the pr!

cmyui (approved):

PR #261, opened 2022-07-24
make old client checks togglable by server operator

0i8:
and also provide the server operator customisability towards how far back they can allow, if they so choose

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -28,6 +28,11 @@ DEBUG=False
 # pages of maps to the official osu! website
 REDIRECT_OSU_URLS=True
 
+# allow/disallow clients older than X days
+# X days defined by CLIENT_TIMEDELTA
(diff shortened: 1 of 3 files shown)


cmyui:
hey haha this is a bit of bad timing

i'm about to push out a commit to rewrite this system to actually go through osu!'s v2 api changelogs and allow any client version up until a major release has been pushed

cmyui:
i suppose we can still implement your feature of allowing the server host to be able to toggle it, but will need to rewrite a bit

cmyui:
i've pushed the new system and opened a pull request here - https://github.com/osuAkatsuki/bancho.py/pull/262

could you move your target branch there and make any changes required?

cmyui (changes requested):
** **

cmyui:
going to implement this myself

image

PR #275, opened 2022-07-30
v4.6.0 - refactor connected service checks & configuration

cmyui:
coding this from 33,000 feet in the sky after a glass of wine btw


cmyui (commented):
lgtm ~goeo_

PR #277, opened 2022-08-04
Maps forced update & More leaderboards & Status check

TrueRou:
1. Add a command and api for BN to force updating a map in order to solve some maps stucking in NotSubmitted or UpdateAvailable status.
2. Add plays and playtime as the sort of leaderboard
3. Change the update logic when maps update available (old ways ignored the condition that a qualified map changed to ranked map without md5 changes)

### DIFF
--- a/app/api/domains/api.py
+++ b/app/api/domains/api.py
@@ -9,14 +9,14 @@ from typing import Optional
 
 import databases.core
 from fastapi import APIRouter
+from fastapi import Header
 from fastapi import HTTPException
 from fastapi import status
-from fastapi import Header
 from fastapi.param_functions import Depends
-from fastapi.security.oauth2 import SecurityScopes
 from fastapi.param_functions import Query
 from fastapi.responses import ORJSONResponse
 from fastapi.responses import StreamingResponse
+from fastapi.security.oauth2 import SecurityScopes
 
 import app.packets
 import app.state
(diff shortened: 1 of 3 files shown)


app/api/domains/api.py
```diff
@@ -938,51 +953,15 @@ async def api_get_pool(
...
-#         ext = "png"
-#     else:
-#         return (400, JSON({"status": "invalid file type."}))
-
-#     # write to the avatar file
-#     (AVATARS_PATH / f"{p.id}.{ext}").write_bytes(ava_file)
-#     return JSON({"status": "success."})
+@router.get("/update_maps")
+async def api_update_maps(api_key: str, sid: int):
+    await check_player(api_key, Privileges.NOMINATOR)
```
tsunyoku, line 958: the flow of calling this function would be very unclear reading this endpoint alone, i’d recommend using Depends() for this or changing the flow to be more obvious

app/objects/beatmap.py
```diff
@@ -737,6 +740,18 @@ async def _update_if_available(self) -> None:
                 {"set_id": self.id},
             )
 
+    async def force_update(self) -> None:
```
tsunyoku, line 743: can we put some space between some of the code here pls, it’s a bit polluted

app/objects/beatmap.py
```diff
@@ -737,6 +740,18 @@ async def _update_if_available(self) -> None:
...
+        await self._update_if_available()
+        app.state.cache.beatmapset.pop(self.id, None)  # drop cache if exist
+        for each_map in self.maps:
+            app.state.cache.beatmap.pop(each_map.md5, None)  # drop cache if exist
+            app.state.cache.beatmap.pop(each_map.id, None)  # drop cache if exist
+            app.state.cache.unsubmitted.discard(each_map.md5)  # drop cache if exist
+            app.state.cache.needs_update.discard(each_map.md5)  # drop cache if exist
+        osu_file_path = BEATMAPS_PATH / f"{each_map.id}.osu"
+        await ensure_local_osu_file(osu_file_path, each_map.id, each_map.md5)
+        await asyncio.sleep(0.5)
```
tsunyoku, line 753: unsure why there’s a sleep here?

app/objects/beatmap.py
```diff
@@ -737,6 +740,18 @@ async def _update_if_available(self) -> None:
                 {"set_id": self.id},
             )
 
+    async def force_update(self) -> None:
+        await self._update_if_available()
```
tsunyoku, line 744: you should check that the map did actually change before deleting cache and/or redownloading the .osu file(s). you should also repopulate the cache in the case it did change

app/objects/beatmap.py
```diff
@@ -737,6 +740,18 @@ async def _update_if_available(self) -> None:
...
+        app.state.cache.beatmapset.pop(self.id, None)  # drop cache if exist
+        for each_map in self.maps:
+            app.state.cache.beatmap.pop(each_map.md5, None)  # drop cache if exist
+            app.state.cache.beatmap.pop(each_map.id, None)  # drop cache if exist
+            app.state.cache.unsubmitted.discard(each_map.md5)  # drop cache if exist
+            app.state.cache.needs_update.discard(each_map.md5)  # drop cache if exist
+        osu_file_path = BEATMAPS_PATH / f"{each_map.id}.osu"
```
tsunyoku, line 751: line 751 and 752 are missing indentation

tsunyoku (changes requested):

tsunyoku:
also in future, please don’t pr from a master branch and also split this into a PR for each change

cmyui:
hey this looks good! no need to close it - we can keep working on it

app/objects/beatmap.py
TrueRou: to update a map forced seems unnecessary because we had the expiration system (2 hours + 5 hours per year).
but the automaticlly check and update map function seems not works in my instance.
i will check and find another solution if i find something weird.

PR #280, opened 2022-08-14
fix: beatmap status changing without md5 changes

TrueRou:
What the problem is:
The ranked status of the map won't update if its md5 is not changed.

When the problem will happen:
A bancho new map changed its status from Qualified to Ranked.
An bancho old map changed its status from Loved to Ranked.

What I did:
Update ranked status if the map is not frozen.


app/objects/beatmap.py
```diff
@@ -665,7 +665,11 @@ async def _update_if_available(self) -> None:
                         bmap._parse_from_osuapi_resp(new_map)
                         updated_maps.append(bmap)
                     else:
-                        # map is the same, make no changes
+                        # map md5 is the same, but status maybe changed (qualified to ranked)
+                        if not old_map.frozen:  # frozen map should not update status
+                            old_map.status = RankedStatus.from_osuapi(
+                                new_map["approved"],
+                            )
```
cmyui, line 672: hey - i don't believe this change will be saved into sql

cmyui (changes requested):
see comment

app/objects/beatmap.py
arily, line 672: there's a `updated_maps.append(old_map)` right after these lines, and 
https://github.com/TrueRou/bancho.py/blob/f5590ff474177cc0d89382211ffc03f288fc5b4e/app/objects/beatmap.py#L719
self._save_to_sql() at the end.
Could you please check again if changes are not reflected to database?

arily: Trou is busy preparing for gaokao test and might not be available for quite a while.

PR #285, opened 2022-08-30
chore(api): change default player count on lb from 25 to 50

tsunyoku:
most places work under the assumption that pages contain 50 players on each page, which also seems more logical to me.

can be closed if too irrelevant but just something i noticed


cmyui:
fine with me, but there are other API endpoints with a limit of 25 at the moment - could you check if those make sense to bump to 50 as well?

tsunyoku:
yeah can do

cmyui (approved):

PR #293, opened 2022-09-14
Fix broken seasonal backgrounds

Sensokaku:
After pulling the last few commits I've noticed seasonal backgrounds don't work and comes up with 

`
File "/home/sensokaku/bancho.py/app/api/domains/osu.py", line 1696, in osuSeasonal                                                                          return ORJSONResponse(app.settings.SEASONAL_BGS._items)
AttributeError: 'list' object has no attribute '_items'
`

And removing _items solved the issue.

I'm not sure if anyone else has gotten the issue (I'm also using ubuntu 22.10 LTS if that could be a factor)

### DIFF
--- a/app/api/domains/osu.py
+++ b/app/api/domains/osu.py
@@ -1650,7 +1650,7 @@ async def osuMarkAsRead(
 
 @router.get("/web/osu-getseasonal.php")
 async def osuSeasonal():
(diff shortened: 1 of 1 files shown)


tsunyoku (approved):
yeah this was due to the config system changing, it's a valid bug

7ez:
cmyui 💀

cmyui (approved):

PR #297, opened 2022-10-16
Spelling mistake

Sukooo:
It should be "its" (possessive pronoun) instead of "it's" because it refers to the score object.


cmyui (approved):

cmyui:
cheers!

PR #316, opened 2022-12-25
add /search_players api route

0i8:
[works well on my server](https://api.halcyon.moe/search?src=j), thought it would be beneficial enough to bring to base bpy

written together with [syswhite](https://github.com/SysWhiteDev)

### DIFF
--- a/app/api/domains/api.py
+++ b/app/api/domains/api.py
@@ -115,14 +115,11 @@ async def api_search_players(
     db_conn: databases.core.Connection = Depends(acquire_db_conn),
 ):
     """Search for users on the server by name."""
-
-    # execute the query using the database connection
     rows = await db_conn.fetch_all(
         "SELECT id, name FROM users WHERE name LIKE :search AND priv & 3 = 3 ORDER BY id ASC",
         {"search": f"%{search}%" if search is not None else None},
     )
 
-    # if the search returns nothing, we tell that to the user.
     if len(rows) == 0:
         return ORJSONResponse(
             {
@@ -131,7 +128,7 @@ async def api_search_players(
             },
             status_code=status.HTTP_404_NOT_FOUND,
         )
-    else:  # else, we return whatever matches we found
+    else:
         return ORJSONResponse(
             {
                 "status": "success",
--- a/app/api/domains/api.py
+++ b/app/api/domains/api.py
@@ -119,7 +119,7 @@ async def api_search_players(
     # execute the query using the database connection
     rows = await db_conn.fetch_all(
         "SELECT id, name FROM users WHERE name LIKE :search AND priv & 3 = 3 ORDER BY id ASC",
-        {"search": f"%{search}%"},
+        {"search": f"%{search}%" if search is not None else None},
     )
 
     # if the search returns nothing, we tell that to the user.
--- a/app/api/domains/api.py
+++ b/app/api/domains/api.py
@@ -6,6 +6,7 @@ import struct
 from pathlib import Path as SystemPath
 from typing import Literal
 from typing import Optional
+from sqlalchemy import text
 
 import databases.core
 from fastapi import APIRouter
@@ -38,6 +39,7 @@ router = APIRouter(tags=["bancho.py API"])
 # or keep up with changes to https://github.com/JKBGL/gulag-api-docs.
 
 # Unauthorized (no api key required)
+# GET /search: returns a list of matching users, based on a passed string, sorted by ascending ID.
 # GET /get_player_count: return total registered & online player counts.
 # GET /get_player_info: return info or stats for a given player.
 # GET /get_player_status: return a player's current status, if online.
@@ -107,6 +109,40 @@ def format_map_basic(m: Beatmap) -> dict[str, object]:
         "diff": m.diff,
     }
 
+@router.get("/search")
+async def api_search(
+    search: Optional[str] = Query(None, alias='src'),
+    db_conn: databases.core.Connection = Depends(acquire_db_conn),
+):
+    """Search for users on the server by name."""
+    
+    # we're using text() here because we are going to
+    # assign placeholders (bindparams()) to these queries, 
+    # and this will let us bind those safely and efficiently.
+    results_query = text("SELECT COUNT(id) FROM users WHERE name AND priv >= 3 LIKE :search_clause") # we will only need to lookup users who have Privileges.VERIFIED
+    rows_query = text("SELECT id, name FROM users WHERE name LIKE :search_clause AND priv >= 3 ORDER BY id ASC") # to save on host resources and prevent useless requests
+
+    # using parameterised queries helps prevent against sql injections
+    # and will also allow the database engine to cache our queries,
+    # making these lookups faster in the future
+    search_clause = f"%{search}%" if search else None
+    results_query = results_query.bindparams(search_clause=search_clause)
+    rows_query = rows_query.bindparams(search_clause=search_clause)
+
+    # execute the queries using the database connection
+    results = await db_conn.fetch_val(results_query)
+    rows = await db_conn.fetch_all(rows_query)
+
+    # return the response
+    response = ORJSONResponse(
+        {
+            "status": "success",
+            "results": results,
+            "result": [dict(row) for row in rows]
+        }
+    )
+
+    return response
 
 @router.get("/get_player_count")
 async def api_get_player_count():
--- a/app/api/domains/api.py
+++ b/app/api/domains/api.py
@@ -118,7 +118,7 @@ async def api_search_players(
 
     # execute the query using the database connection
     rows = await db_conn.fetch_all(
-        "SELECT id, name FROM users WHERE name LIKE :search AND priv >= 3 ORDER BY id ASC",
+        "SELECT id, name FROM users WHERE name LIKE :search AND priv & 3 = 3 ORDER BY id ASC",
         {"search": f"%{search}%"},
     )
 
--- a/app/api/domains/api.py
+++ b/app/api/domains/api.py
@@ -118,9 +118,9 @@ async def api_search_players(
 
     # execute the query using the database connection
     rows = await db_conn.fetch_all(
-        "SELECT id, name FROM users WHERE name LIKE :search AND priv >= 3 ORDER BY id ASC", 
-        {"search": f"%{search}%"}
-        )
+        "SELECT id, name FROM users WHERE name LIKE :search AND priv >= 3 ORDER BY id ASC",
+        {"search": f"%{search}%"},
+    )
 
     # if the search returns nothing, we tell that to the user.
     if len(rows) == 0:
--- a/app/api/domains/api.py
+++ b/app/api/domains/api.py
@@ -112,7 +112,7 @@ def format_map_basic(m: Beatmap) -> dict[str, object]:
 
 @router.get("/search")
 async def api_search(
-    search: Optional[str] = Query(None, alias="q"),
+    search: Optional[str] = Query(None, alias="q", min=2, max=32),
     db_conn: databases.core.Connection = Depends(acquire_db_conn),
 ):
     """Search for users on the server by name."""
--- a/app/api/domains/api.py
+++ b/app/api/domains/api.py
@@ -120,22 +120,13 @@ async def api_search_players(
         {"search": f"%{search}%" if search is not None else None},
     )
 
-    if len(rows) == 0:
-        return ORJSONResponse(
-            {
-                "status": "error",
-                "message": "nothing was found matching those parameters!",
-            },
-            status_code=status.HTTP_404_NOT_FOUND,
-        )
-    else:
-        return ORJSONResponse(
-            {
-                "status": "success",
-                "results": len(rows),
-                "result": [dict(row) for row in rows],
-            },
-        )
+    return ORJSONResponse(
+        {
+            "status": "success",
+            "results": len(rows),
+            "result": [dict(row) for row in rows],
+        },
+    )
 
 
 @router.get("/get_player_count")
--- a/app/api/domains/api.py
+++ b/app/api/domains/api.py
@@ -141,12 +141,19 @@ async def api_search(
     # if the search returns nothing, we tell that to the user.
     if results < 1:
         return ORJSONResponse(
-            {"status": "error", "message": "nothing was found matching those parameters!"},
+            {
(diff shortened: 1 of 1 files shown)


app/api/domains/api.py
```diff
@@ -121,7 +121,7 @@ async def api_search(
     # assign placeholders (bindparams()) to these queries,
     # and this will let us bind those safely and efficiently.
     results_query = text(
```
minisbett, line 123: You also don't need this whole query at all, you can just all len() on the resulting list of users.

minisbett (changes requested):

app/api/domains/api.py
```diff
@@ -107,6 +109,40 @@ def format_map_basic(m: Beatmap) -> dict[str, object]:
         "diff": m.diff,
     }
 
+@router.get("/search")
+async def api_search(
+    search: Optional[str] = Query(None, alias='src'),
```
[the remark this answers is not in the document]
7ez: "query" or "q" would be more preferable imo

app/api/domains/api.py
```diff
@@ -107,6 +109,40 @@ def format_map_basic(m: Beatmap) -> dict[str, object]:
...
+    search_clause = f"%{search}%" if search else None
+    results_query = results_query.bindparams(search_clause=search_clause)
+    rows_query = rows_query.bindparams(search_clause=search_clause)
+
+    # execute the queries using the database connection
+    results = await db_conn.fetch_val(results_query)
+    rows = await db_conn.fetch_all(rows_query)
+
+    # return the response
+    response = ORJSONResponse(
```
7ez: why are you assigning the response instead of just retuurning it?

app/api/domains/api.py
```diff
@@ -107,6 +109,40 @@ def format_map_basic(m: Beatmap) -> dict[str, object]:
...
+    # assign placeholders (bindparams()) to these queries, 
+    # and this will let us bind those safely and efficiently.
+    results_query = text("SELECT COUNT(id) FROM users WHERE name AND priv >= 3 LIKE :search_clause") # we will only need to lookup users who have Privileges.VERIFIED
+    rows_query = text("SELECT id, name FROM users WHERE name LIKE :search_clause AND priv >= 3 ORDER BY id ASC") # to save on host resources and prevent useless requests
+
+    # using parameterised queries helps prevent against sql injections
+    # and will also allow the database engine to cache our queries,
+    # making these lookups faster in the future
+    search_clause = f"%{search}%" if search else None
+    results_query = results_query.bindparams(search_clause=search_clause)
```
7ez, line 134: why not let the database handle parameter binding?

app/api/domains/api.py
0i8: this was already changed in the last commit

minisbett: Check out the code I sent in general on the Discord, I'd suggest replacing this whole PR with that code

7ez: my code* :trollface:

minisbett: https://discord.com/channels/748687781605408908/748687781605408911/1056576143668555907

minisbett: > my code* :trollface:

Didn't say it's mine

app/api/domains/api.py
0i8: using bindparams() lets me bind the values safely, for when i am constructing queries dynamically (such as in this route). this is the safest way to avoid SQL injection, and improves performance on search because the engine will cache our queries.

in contrast, concatenating values directly into the query string can open us up for SQL injection, and makes our queries slower and uncacheable. if you have a different approach that avoids the latter, i would love to hear it

app/api/domains/api.py
0i8: no

app/api/domains/api.py
minisbett: Check the code on the server...

app/api/domains/api.py
0i8: len() is slower by 50+ms on every query if i used it. 

count() is more efficient because it returns the number of rows in the condition specified without having to retrieve all of the rows after that query is done, and then count those rows.

app/api/domains/api.py
7ez: you can just do
```py
await db_conn.fetch_val("SELECT name FROM users WHERE id = :id", {"id":1})
```
for example

app/api/domains/api.py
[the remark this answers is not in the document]
0i8: ok

app/api/domains/api.py
7ez: > len() is slower by 50+ms on every query if i used it. 
> 
> count() is more efficient because it returns the number of rows in the condition specified without having to retrieve all of the rows after that query is done, and then counting those rows.

idk what you're talking about, I ran a timeit test on `len([1, 2, 3] * 100)` and got the result of ```2000000 loops, best of 5: 6.26 usec per loop```

app/api/domains/api.py
0i8: i don't agree with this approach, because in most cases with this route, we're executing multiple queries handled by a frontend where `fetch_val()` is slower overall than `bindparams()`

if we're going to be dynamically constructing a query, then `bindparams()` is the way i want to do this.
if we are just going to be executing a singular query with fixed values, then i would agree with `fetch_val()`

7ez: > i don't agree with this approach, because in most cases with this route, we're executing multiple queries handled by a frontend where `fetch_val()` is slower overall than `bindparams()`
> 
> if we're going to be dynamically constructing a query, then `bindparams()` is the way i want to do this.
> if we are just going to be executing a singular query with fixed values, then i would agree with `fetch_val()`
> 
> both could be equally good towards this, i would like some more input before i make any changes though

fetch_val was an example. It works on any fetch or execute method.

minisbett: Then rewrite all of bpy so everything is uniform again

app/api/domains/api.py
0i8: the implementation of len() for this wouldn't work on my machine and only returned no values every time, idk what it thinks its doing lol

app/api/domains/api.py
cmyui: hey, the databases library were using is already doing this stuff behind the scenes for you - no need to use it directly here

app/api/domains/api.py
```diff
@@ -38,6 +39,7 @@
 # or keep up with changes to https://github.com/JKBGL/gulag-api-docs.
 
 # Unauthorized (no api key required)
+# GET /search: returns a list of matching users, based on a passed string, sorted by ascending ID.
```
cmyui, line 42: ideally, this should be something like `GET /players`, but since the whole API is currently wrong, let's keep it consistently wrong 🥲 
```suggestion
# GET /get_players: returns a list of matching users, based on a passed string, sorted by ascending ID.
```

app/api/domains/api.py
```diff
@@ -108,6 +110,53 @@ def format_map_basic(m: Beatmap) -> dict[str, object]:
     }
 
 
+@router.get("/search")
```
cmyui, line 113:
```suggestion
@router.get("/get_players")
```

app/api/domains/api.py
```diff
@@ -108,6 +110,53 @@ def format_map_basic(m: Beatmap) -> dict[str, object]:
...
+async def api_search(
```
cmyui, line 114:
```suggestion
async def api_search_players(
```

app/api/domains/api.py
cmyui: this puts me thru great pain

app/api/domains/api.py
```diff
@@ -108,6 +110,53 @@ def format_map_basic(m: Beatmap) -> dict[str, object]:
...
+
+    # we're using text() here because we are going to
+    # assign placeholders (bindparams()) to these queries,
+    # and this will let us bind those safely and efficiently.
+    results_query = text(
+        "SELECT COUNT(id) FROM users WHERE name LIKE :search_clause AND priv >= 3",
+    )  # we will only need to lookup users who have Privileges.VERIFIED
+    rows_query = text(
+        "SELECT id, name FROM users WHERE name LIKE :search_clause AND priv >= 3 ORDER BY id ASC",
+    )  # to save on host resources and prevent useless requests
```
cmyui, line 128: bit more correct
```suggestion
        "SELECT COUNT(id) FROM users WHERE name LIKE :search_clause AND priv & 3 = 3",
    )  # we will only need to lookup users who have Privileges.VERIFIED
    rows_query = text(
        "SELECT id, name FROM users WHERE name LIKE :search_clause AND priv & 3 = 3 ORDER BY id ASC",
    )  # to save on host resources and prevent useless requests
```

cmyui:
aw man, this just reminds me how bad the current API is

app/api/domains/api.py
0i8: sure, we'll use it

app/api/domains/api.py
```diff
@@ -108,6 +109,38 @@ def format_map_basic(m: Beatmap) -> dict[str, object]:
...
+async def api_search_players(
+    search: Optional[str] = Query(None, alias="q", min=2, max=32),
+    db_conn: databases.core.Connection = Depends(acquire_db_conn),
+):
+    """Search for users on the server by name."""
+
+    # execute the query using the database connection
+    rows = await db_conn.fetch_all(
+        "SELECT id, name FROM users WHERE name LIKE :search AND priv & 3 = 3 ORDER BY id ASC",
+        {"search": f"%{search}%"},
```
cmyui, line 122: not sure this handles the None case correctly at the moment - i think we'd want 
```suggestion
        {"search": f"%{search}%" if search is not None else None},
```

cmyui: if you could test this, would be epic

app/api/domains/api.py
```diff
@@ -108,6 +109,38 @@ def format_map_basic(m: Beatmap) -> dict[str, object]:
...
 
 
+@router.get("/search_players")
+async def api_search_players(
+    search: Optional[str] = Query(None, alias="q", min=2, max=32),
+    db_conn: databases.core.Connection = Depends(acquire_db_conn),
+):
+    """Search for users on the server by name."""
+
+    # execute the query using the database connection
```
cmyui, line 119: code is self explanatory here
```suggestion
```

app/api/domains/api.py
cmyui: maybe we can just disallow empty query; might be simpler

app/api/domains/api.py
cmyui, line 119: i used to put comments like this all over my code, but my new mindset is https://www.youtube.com/watch?v=2a_ytyt9sf8&t=626s

cmyui: the code here is actually very self-expressive though, i think it's quite readable without the comments

0i8: while i don't agree with this mindset, i will apply it here

cmyui: would recommend actually watching into the vid though, he explains it more in-depth - e.g. refactoring your code to read *as if it were a comment*;
image

app/api/domains/api.py
```diff
@@ -108,6 +109,35 @@ def format_map_basic(m: Beatmap) -> dict[str, object]:
...
+        "SELECT id, name FROM users WHERE name LIKE :search AND priv & 3 = 3 ORDER BY id ASC",
+        {"search": f"%{search}%" if search is not None else None},
+    )
+
+    if len(rows) == 0:
+        return ORJSONResponse(
+            {
+                "status": "error",
+                "message": "nothing was found matching those parameters!",
+            },
```
cmyui, line 128: hmm, seems this is inconsistent with my old style - scrolling down i see some like
```py
return ORJSONResponse(
    {"status": "Invalid gamemode."},
    status_code=status.HTTP_400_BAD_REQUEST,
)
```
i actually prefer what you have lol, it's what i use in my newer projects

quite unfortunate

app/api/domains/api.py
0i8: this makes more sense, i'll watch the video when i get a chance

app/api/domains/api.py
0i8: pre-commit changed that, originally it was in line with what you have already

cmyui, line 128:
```suggestion
            {"status": "Nothing was found matching those parameters."},
```

cmyui: actually, on second thought, let's return 200 with [] result

app/api/domains/api.py
```diff
@@ -108,6 +109,35 @@ def format_map_basic(m: Beatmap) -> dict[str, object]:
...
+            status_code=status.HTTP_404_NOT_FOUND,
+        )
+    else:
+        return ORJSONResponse(
+            {
+                "status": "success",
+                "results": len(rows),
+                "result": [dict(row) for row in rows],
+            },
+        )
```
cmyui, line 138:
```suggestion
    return ORJSONResponse(
        {
            "status": "success",
            "results": len(rows),
            "result": [dict(row) for row in rows],
        },
    )
```

cmyui (changes requested):
need to test null case before i can merge this

app/api/domains/api.py
0i8: why always return 200?

7ez: no point in doing more serverside work

cmyui: this is more standard in rest

when fetching an array of resources, an empty array is still a valid response
an example might be, e.g. `GET /users` returning `200 OK` with body `[]`

404 is more for the case where you're looking up a specific resource and it is not found
an example might be, e.g. `GET /users/1` returning `404 NOT FOUND`

0i8:
> need to test null case before i can merge this

what is the intended behaviour? looks and works fine in all cases it appears, on my end

cmyui:
> > need to test null case before i can merge this
> 
> what is the intended behaviour? looks and works fine in all cases it appears, on my end

just wasn't sure the null case would work - i've got a setup running now, i'll get to this shortly np

cmyui:
returns all players upon null query now, as i'd expect

cmyui (approved):
+1 👍🏻

0i8:
img

PR #317, opened 2022-12-28
refactor: mods.py overhaul

g1-1-1:
starting from the beginning, detailing everything i changed and why:

\_\_repr\_\_: this method is made more efficient because it iterates over the `Mods` enum values and checks if each value is set in the `Mods` object. if it is set, the corresponding mod string is added to the `mod_str` list. this is more efficient than the previous version because it avoids unnecessary string manipulation and only performs a single iteration over the enum values of `Mods`.

filter_invalid_combos: instead of using a series of if statements to check for conflicting mod combinations, we could use a dictionary to map the conflicting mod combinations to their resolution. now we only need to check for conflicting mod combinations once, instead of for each if statement. this is also easier to read and maintain, as the mapping between conflicting mod combinations and their resolution is explicit.

from_modstr: removed the unnecessary check for `m` not being in `_dict`, since this will always be true if `m` is in `mod_strs`. i also changed the continue statement to simply not add the mod to mods if `m` is not in `_dict`. this is equivalent to the original code, but simpler and more readable.

from_np: this refactored version of the method uses a dictionary comprehension to create a dictionary that maps np strings to mod values, which allows us to more efficiently look up the mod value for a given np string. it also splits the input string into a list of mod strings and iterates over that list, rather than iterating over the words in the input string and calling split on each iteration.

as always i would love feedback, and proper testing -- i wrote it and briefly tested it and it seemed to work but i need more eyes on it


app/constants/mods.py
```diff
@@ -53,113 +54,96 @@ def __repr__(self) -> str:
...
-            # keep only the first
-            first_keymod = None
-            for mod in KEY_MODS:
-                if keymods_used & mod:
-                    first_keymod = mod
-                    break
-
-            # remove all but the first keymod.
+            # keep only the first keymod
+            first_keymod = 2 ** int(math.log(keymods_used, 2))
```
cmyui, line 115: idk how this works LOL

app/constants/mods.py
```diff
@@ -53,113 +54,96 @@ def __repr__(self) -> str:
...
+                Mods.AUTOPILOT,
+                Mods.PERFECT,
+            ): Mods.NOFAIL,  # (NF|RX|AP)PF
+            (Mods.RELAX, Mods.AUTOPILOT, Mods.NOFAIL): Mods.RELAX,  # (RX|AP)NF
+            (Mods.PERFECT, Mods.SUDDENDEATH): Mods.PERFECT,  # PFSD
+            (Mods.HIDDEN, Mods.FADEIN): Mods.HIDDEN,  # HDFI
+        }
+
+        # check for conflicting mod combinations and apply the resolution
+        for conflict, resolution in conflicts.items():
```
cmyui, line 89: the name "resolution" to me sounded more like the end result, but it's actually what's removed from the end result

seems a bit counterintuitively named

app/commands.py
```diff
@@ -344,7 +344,7 @@ async def recent(ctx: Context) -> Optional[str]:
     l = [f"[{s.mode!r}] {s.bmap.embed}", f"{s.acc:.2f}%"]
 
     if s.mods:
-        l.insert(1, f"+{s.mods!r}")
+        l.insert(1, f"+{Mods.filter_invalid_combos(s.mods, s.mode)!r}")
```
cmyui, line 347: this is from an existing score, so it shouldn't be able to get into the db in an invalid state

once it's in db, i trust it, so don't need to call this (expensive) func on it

cmyui:
overall tbh i think this just increases complexity

if statements are dead simple, first thing people learn in programming, i think the dict approach is more modular (it abstracts it into more of a config, which is certainly valuable in some scenarios) but still increases complexity and i don't really think there's much benefit to having it configurable - invalid mod combos are a pretty globally agreed-upon subject (generally it's just things the osu! client doesn't allow together)

app/commands.py
```diff
@@ -538,7 +538,7 @@ async def _with(ctx: Context) -> Optional[str]:
 
     # include mods regardless of mode
     if (mods := command_args.get("mods")) is not None:
-        msg_fields.append(f"{mods!r}")
+        msg_fields.append(f"{Mods.filter_invalid_combos(mods, mode_vn)!r}")
```
cmyui, line 541: this is probably correct

cmyui:
if you're looking to optimize this, the real big thing that will increase performance by a couple of orders of magnitude is to stop using the enum module IntFlag instances for Mods and to use bare ints - iirc, filter_invalid_combos runs on the order of microseconds because the bitwise operations run something like 100-1000 python operation in the background. making this an int will move this down to 1-2 operations

app/constants/mods.py
0i8: it makes sense to me because the removal of the invalid combination is the resolution, but i can get why someone could get confused about it

i got no idea how to name it though

cmyui:
@tsunyoku thoughts?

app/commands.py
0i8: sure that makes sense

cmyui:
> the real big thing that will increase performance by a couple of orders of magnitude is to stop using the enum module IntFlag instances for Mods and to use bare ints

i should add, this is possible while still using a class namespace to contain the enums; something like

```py
class Mods:
    A = 1 << 0
    B = 1 << 1
    ...

Mods.A | Mods.B
```
but there are some things you lose, e.g. the __repr__ functionality that the enum module classes provide. I tried to make my own enum implementation back in the day when I was working (& trying to improve performance) on bancho.py back in 2020/2021, but never got around to it as there were always higher priorities

app/constants/mods.py
```diff
@@ -53,113 +54,96 @@ def __repr__(self) -> str:
...
-                mod_str.append(_dict[mod])
-
+        mod_str = [mod2modstr_dict[mod] for mod in Mods if self.value & mod]
         return "".join(mod_str)
 
     def filter_invalid_combos(self, mode_vn: int) -> Mods:
         """Remove any invalid mod combinations."""
-
-        # 1. mode-inspecific mod conflictions
-        _dtnc = self & (Mods.DOUBLETIME | Mods.NIGHTCORE)
```
cmyui, line 69: note that i was previously only doing these types of things because of how expensive the IntFlag bitwise operations are as well - i was really trying to squeeze performance out wherever i could lol

cmyui: with bare ints, i wouldn't even bother as this hurts readability quite a bit (in bare ints, this would probably be ~100 nanoseconds)

0i8:
> > the real big thing that will increase performance by a couple of orders of magnitude is to stop using the enum module IntFlag instances for Mods and to use bare ints
> > I'd be open to reviewing a PR doing something like this (preferably with the other classes also refactored, like Privileges as well for consistency)

while i'm open to refactoring the mods class like this i don't think i could fathom doing it with privileges, mostly because they're referenced too often throughout the code and i'm a lazy sonova bitch

cmyui:
> while i'm open to refactoring the mods class like this i don't think i could fathom doing it with privileges, mostly because they're referenced too often throughout the code and i'm a lazy sonova bitch

you won't have to touch most of the references, only the times where we're using the functionality specific to the IntFlag class, or the functions we've added to the class namespace
`Mods.__repr__()` -> `privileges.to_string(Mods())`
`Mods.filter_invalid_combos` -> `mods.filter_invalid_combos(Mods())`
and probably a few others

from taking a quick look, i see 193 references to Privileges, with probably only <10% of those needing edits
Mods should be the same. I can help with these too, if you get the ball rolling.
the nice thing with git is you can work gradually in branches; we don't have to do this all overnight

cmyui:
still interested in @tsunyoku thoughts though

0i8:
can we have a final pass on this? i am not sure if i missed anything/i did anything wrong

app/constants/mods.py
```diff
@@ -48,228 +39,194 @@ class Mods(IntFlag):
...
-                if keymods_used & mod:
-                    first_keymod = mod
-                    break
+        keymods_used = mods & KEY_MODS
 
-            # remove all but the first keymod.
-            self &= ~(keymods_used & ~first_keymod)
+        if keymods_used & (keymods_used - 1):
+            # remove all but the lowest keymod
+            mods &= keymods_used & -keymods_used
```
cmyui, line 129: hey, how does this work? still not sure

0i8: my theory (which works), as that `keymods_used & (keymods_used - 1)` checks if there are multiple keymods enabled at the same time by first ANDing `keymods_used` with `keymods_used - 1`
if there are multiple keymods enabled; `keymods_used` will have multiple bits set to 1, so `keymods_used - 1` will have at least one bit set to 1 that is not set in `keymods_used`. when these values are ANDed together, the result will be a non-zero value, indicating that multiple keymods are enabled. 

if only one keymod is enabled, `keymods_used - 1` will have all bits set to 0, so the result of `keymods_used & (keymods_used - 1)` will be 0, indicating that only one keymod is enabled, otherwise; it will execute what's in the if statement.

this appears to work flawlessly on my machine but still needs more testing from other parties to be sure

cmyui:
hey i'm not particularly fond of the changes unrelated to removing the IntFlag part (changing bitwise op orders, comments, constants like OSU_SPECIFIC_MODS/SCORE_INCREASE_MODS/etc.)

could those get reverted? i'm really just interested in removing the IntFlag, the rest can come in separate proposals

0i8:
> changing bitwise op orders
i didn't do this

> comments
the comments i removed were redundant since the code is explaining itself

> OSU_SPECIFIC_MODS/SCORE_INCREASE_MODS
these weren't being used at the moment since we already filter out by mode using filter_invalid_combos(), other than that i'm not sure of a potential usecase where we need to call these specifically and they weren't being presently used anywhere

i'm not sure what was wanted here since the PR name suggests that i planned to redo this entire class rather than just removing the intflag

tsunyoku (changes requested):
agree with @cmyui, i'm really quite confused as to why some of the changes here exist. only interest should really be in removing IntFlag and ensuring any **necessary** changes as a result of that are made. the readability on the mods.py file with this pr is a major step down

g1-1-1:
i'll probably redo this another time, i don't agree with my approach i took on this

PR #322, opened 2022-12-29
Create and close ClientSession when recalculating

7ez:
As of right now, if `ensure_local_osu_file` can't find a beatmap, it will fail to potentially grab it from osu! as the http client was not created.

### DIFF
--- a/tools/recalc.py
+++ b/tools/recalc.py
@@ -30,7 +30,7 @@ try:
     from app.constants.gamemodes import GameMode
     from app.objects.beatmap import ensure_local_osu_file
     import app.settings
-    import app.state
+    import app.state.services
 except ModuleNotFoundError:
     print("\x1b[;91mMust run from tools/ directory\x1b[m")
     raise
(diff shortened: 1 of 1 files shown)


tools/recalc.py
```diff
@@ -29,6 +30,7 @@
     from app.constants.gamemodes import GameMode
     from app.objects.beatmap import ensure_local_osu_file
     import app.settings
+    import app.state
```
cmyui, line 33: i think this should still work okay, just limits import scope a bit
```suggestion
    import app.state.service
```

cmyui:
heh this reminds me, i need to rip that `ensure_local_osu_file` into like 3 functions

cmyui:
also i suppose i can probably replace aiohttp with httpx

tools/recalc.py
7ez: done

cmyui (approved):
+1 👍🏻

PR #324, opened 2022-12-31
feat: add requested status, comments and declining to ranking process

minisbett:
This modifies the map_requests table and adds a requested_status and comment column.
The !request command now accepts rank(ed)/love(d) as the first argument and optionally a comment as the second.

Additionally the !requests command now displays the requests in a better way and shows the requested status and comment.
The !declinerequest command can be used to deactivate a request without having to change the status.

I feel like I might have messed something up so better re-check everything


7ez:
why would you want a comment

minisbett:
> why would you want a comment

So people can explain why they think it's suitable for ranked and loved? Imagine loved.sh or the discussion section on bancho wouldn't be a thing

7ez:
> > why would you want a comment
> 
> So people can explain why they think it's suitable for ranked and loved? Imagine loved.sh or the discussion section on bancho wouldn't be a thing 

wouldn't that be up to the bns not the users

minisbett:
> > > why would you want a comment
> > 
> > 
> > So people can explain why they think it's suitable for ranked and loved? Imagine loved.sh or the discussion section on bancho wouldn't be a thing
> 
> wouldn't that be up to the bns not the users

Yeah but the users can leave a message for the BNs on why they think the map is worth ranking or loving it

0i8:
> > > > why would you want a comment
> > > 
> > > 
> > > So people can explain why they think it's suitable for ranked and loved? Imagine loved.sh or the discussion section on bancho wouldn't be a thing
> > 
> > 
> > wouldn't that be up to the bns not the users
> 
> Yeah but the users can leave a message for the BNs on why they think the map is worth ranking or loving it

i've had this implemented for a while and i've seen it used like twice in the 200+ requests i've had

app/commands.py
```diff
@@ -543,11 +546,13 @@ async def request(ctx: Context) -> Optional[str]:
     if bmap.status != RankedStatus.Pending:
         return "Only pending maps may be requested for status change."
 
+    status = "ranked" if ctx.args[0] == "rank" or ctx.args[0] == "ranked" else "loved";
```
cmyui, line 549: HOLY rare python semicolon

migrations/migrations.sql
```diff
@@ -402,3 +402,8 @@ insert into achievements (id, file, name, `desc`, cond) values (83, 'all-intro-s
 
 # v4.4.3
 alter table favourites add created_at int default 0 not null;
+
+# v4.6.5
+alter table map_requests add requested_status varchar(6) default '' not null;
+alter table map_requests add comment text default '' not null;
```
cmyui, line 408:
```suggestion
alter table map_requests add requested_status varchar(6) default '' not null after map_id;
alter table map_requests add comment text default '' not null after requested_status ;
```

cmyui: mysql lets us specify order, unlike psql

app/commands.py
```diff
@@ -543,11 +546,13 @@ async def request(ctx: Context) -> Optional[str]:
...
+    
     await app.state.services.database.execute(
         "INSERT INTO map_requests "
-        "(map_id, player_id, datetime, active) "
-        "VALUES (:map_id, :user_id, NOW(), 1)",
-        {"map_id": bmap.id, "user_id": ctx.player.id},
+        "(map_id, requested_status, comment, player_id, datetime, active) "
+        "VALUES (:map_id, :status, :comment, :user_id, NOW(), 1)",
+        {"map_id": bmap.id, "status": status, "comment": "".join(ctx.args[1:]), "user_id": ctx.player.id},
```
cmyui, line 555: i think we should limit the length of the comment.

not sure if the message has already been truncated during packet handling, i think so? if i remember correctly, i truncate to ~2048 or so

app/commands.py
```diff
@@ -611,10 +616,31 @@ async def requests(ctx: Context) -> Optional[str]:
...
+                 + f"{requested_status} by {p.embed}. "
+                 + f"{'No additional comment was specified.' if comment == '' else f'Additional comment: {comment}'}")
 
     return "\n".join(l)
 
+@command(Privileges.NOMINATOR, aliases=["dr"], hidden=True)
+async def declinerequest(ctx: Context) -> Optional[str]:
+    """Removes a map request by it's ID."""
+    if len(ctx.args) != 1:
+        return "Invalid syntax: !declinerequest <ID>"
```
cmyui, line 629: consistent with !mp map
```suggestion
        return "Invalid syntax: !declinerequest <beatmapid>"
```

app/commands.py
```diff
@@ -611,10 +616,31 @@ async def requests(ctx: Context) -> Optional[str]:
...
+
+    if await app.state.services.database.fetch_one(
+        "SELECT 1 FROM map_requests WHERE active = 1 AND id = :id",
+        {"id": ctx.args[0]}
+    ):
+        return f"No active map request with ID #{ctx.args[0]} found."
```
cmyui, line 635: is this missing an `is None` check? perhaps i'm misreading

app/commands.py
```diff
@@ -611,10 +616,31 @@ async def requests(ctx: Context) -> Optional[str]:
...
             continue
 
-        l.append(f"[{p.embed} @ {dt:%b %d %I:%M%p}] {bmap.embed}.")
+        l.append(f"#{id}: {bmap.embed}\n({dt:%b %d %I:%M%p}) Requested for "
+                 + f"{requested_status} by {p.embed}. "
+                 + f"{'No additional comment was specified.' if comment == '' else f'Additional comment: {comment}'}")
 
     return "\n".join(l)
 
+@command(Privileges.NOMINATOR, aliases=["dr"], hidden=True)
```
cmyui, line 625: seems like nice middleground
```suggestion
@command(Privileges.NOMINATOR, aliases=["decline", "dr"], hidden=True)
```

app/commands.py
cmyui: i wonder if this makes more sense to be based on /np though?

cmyui (commented):
nice work - not sure how much this will be used, but it's certainly an improvement on the existing system nevertheless.

left some thoughts - let me know what you think.

app/commands.py
minisbett: lmfao

app/commands.py
minisbett: From when I tried this on my server I felt like it was more handy to just put the ID there from the request so you wouldn't even have to download some maps, I feel like this is easier than nping

app/commands.py
minisbett: Got this if check from somewhere else, isn't if ... and if ... is not None the same?

app/commands.py
0i8: maybe give the nominator the option to do both if they want?

app/commands.py
```diff
@@ -532,8 +532,15 @@ async def _with(ctx: Context) -> Optional[str]:
...
-    if ctx.args:
-        return "Invalid syntax: !request"
+    if (
+        len(ctx.args) < 1
+        or ctx.args[0] not in ("rank", "ranked", "love", "loved")
+    ):
+        return "Invalid syntax: !request <ranked/loved> [optional comment]"
+    
+    comment = "".join(ctx.args[1:])
+    if(len(ctx.args) > 1 && len(comment) > 512)
```
cmyui, line 542:
```suggestion
    if len(ctx.args) > 1 and len(comment) > 512:
```

minisbett: lmfao

app/commands.py
```diff
@@ -612,11 +637,56 @@ async def requests(ctx: Context) -> Optional[str]:
             l.append(f"Failed to find requested map ({map_id})?")
             continue
 
-        l.append(f"[{player.embed} @ {dt:%b %d %I:%M%p}] {bmap.embed}.")
+        l.append(
+            f"#{id}: {bmap.embed}\n({dt:%b %d %I:%M%p}) Requested for "
+            + f"{requested_status} by {p.embed}. "
```
tsunyoku, line 642: `requested_status` could be None, this should be handled

app/commands.py
```diff
@@ -612,11 +637,56 @@ async def requests(ctx: Context) -> Optional[str]:
...
+            + (f"Additional comment: {comment}" if comment else ""),
+        )
 
     return "\n".join(l)
 
 
+@command(Privileges.NOMINATOR, aliases=["dr", "decline"], hidden=True)
+async def declinerequest(ctx: Context) -> Optional[str]:
```
tsunyoku, line 650: i'm unsure about this function as a whole regarding the fact it clears **all** requests. there could be a ranked request that you want to deny but not deny the loved one. perhaps declining by request ID should be an enforced behaviour? curious to hear other's thoughts on this

tsunyoku (changes requested):

migrations/migrations.sql
```diff
@@ -403,9 +403,13 @@ insert into achievements (id, file, name, `desc`, cond) values (83, 'all-intro-s
 # v4.4.3
 alter table favourites add created_at int default 0 not null;
 
+# v4.6.5
```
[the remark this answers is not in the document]
7mochi: @minisbett is this fixed or has it been tested to see if it works as is?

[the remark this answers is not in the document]
minisbett: Yeah this needs to be bumped to the newest bpy version since this PR is ages old

app/commands.py
```diff
@@ -534,8 +534,8 @@ async def _with(ctx: Context) -> Optional[str]:
 @command(Privileges.UNRESTRICTED, aliases=["req"])
 async def request(ctx: Context) -> Optional[str]:
     """Request a beatmap for nomination."""
-    if ctx.args:
-        return "Invalid syntax: !request"
+    if len(ctx.args) < 1:
+        return "Invalid syntax: !request <ranked/loved> [optional comment]"
```
cmyui, line 538: re. `<ranked/loved>`, i think there's limited value in having users able to request a status, i think having the bn team decide will be fine for most servers.

on akatsuki i've always thought it's best to keep the input unbiased and just provide the bn with maps to look at - where they decide it should be categorized (ranked, loved, etc.) is ultimately up to the team imo, so i'm also not sure of the value of `comment`.

i think something more valuable could be having the the team be able to provide feedback for why a request was declined, such that in those cases the downstream users who made the requests have some understanding of how to better nominate maps in the future

minisbett:
deemed not that useful

PR #325, opened 2023-01-02
feat: add base IRC server implementation

7mochi:
Sources that helped me to do this:
  - https://github.com/osu-Sakuru/tamatsu/blob/09df2db73b0c938fd9c4de79cab572a152518e9f/app/irc.py
  - https://github.com/lenforiee/banchoirc_test/blob/master/irc_test.py
  - https://github.com/osuAkatsuki/bancho-service/blob/main/irc/ircserver.py


cmyui:
mostly looks good; i'll take over the branch to do some typing fixes

app/irc.py
```diff
@@ -314,34 +364,43 @@ async def send_message(self, fro: Player, to: str, message: str) -> int:
...
     def bancho_message(self, fro: str, to: str, message: str) -> None:
         if to.startswith("#"):
             for client in self.authorized_clients:
-                if client.player.name != fro:
-                    to in [x.name for x in client.player.channels]
+                assert client.player is not None
+
+                if client.player.name != fro and to in [
+                    x.name for x in client.player.channels
+                ]:
```
cmyui, line 398: @7mochi could you confirm if this is correct?

there was a bug here with `to in [x.name for x in client.player.channels]` not being used in the `if` statement. I assumed this was the correct fix.

7mochi: yep 👍

NiceAesth:
Is this PR still being looked at?

7mochi:
> Is this PR still being looked at?

@lenforiee will implement the rest i think

PR #329, opened 2023-01-07
fix: various beatmap bugs

cmyui:
kudos to @JKBGL for the bug reports

### DIFF
--- a/app/objects/beatmap.py
+++ b/app/objects/beatmap.py
@@ -380,7 +380,15 @@ class Beatmap:
 
             if beatmap_set is not None:
                 # the beatmap set has been cached - fetch beatmap from cache
-                bmap = await cls._from_md5_cache(md5, check_updates=False)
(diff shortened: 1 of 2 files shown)


app/objects/beatmap.py
```diff
@@ -654,7 +648,13 @@ async def _update_if_available(self) -> None:
...
                 else:
                     new_map = new_maps[old_id]
-                    if old_map.md5 != new_map["file_md5"]:
+                    new_ranked_status = RankedStatus.from_osuapi(
+                        int(new_map["approved"]),
+                    )
+                    if (
+                        old_map.md5 != new_map["file_md5"]
+                        or old_map.status != new_ranked_status
+                    ):
```
cmyui, line 657: this was breaking in the case where a qualified map was updated to ranked without a change in md5. map would remain qualified indefinitely.

PR #334, opened 2023-01-07
Make fetching maps from any server at once possible

def750:


app/repositories/maps.py
```diff
@@ -122,7 +122,7 @@ async def create(
 
 
 async def fetch_one(
-    server: str,
+    server: Optional[str] = "osu!",
```
cmyui, line 125:
```suggestion
    server: Optional[str] = None,
```

app/repositories/maps.py
```diff
@@ -186,7 +186,7 @@ async def fetch_count(
 
 
 async def fetch_many(
-    server: str,
+    server: Optinal[str] = "osu!",
```
cmyui, line 189:
```suggestion
    server: Optinal[str] = None,
```

app/repositories/maps.py
```diff
@@ -201,7 +201,7 @@ async def fetch_many(
     query = f"""\
         SELECT {READ_PARAMS}
           FROM maps
-         WHERE server = :server
+         WHERE COALESCE(:server, server)
```
cmyui, line 204:
```suggestion
         WHERE server = COALESCE(:server, server)
```

app/repositories/maps.py
```diff
@@ -186,7 +186,7 @@ async def fetch_count(
 
 
 async def fetch_many(
-    server: str,
+    server: Optinal[str] = None,
```
cmyui, line 189:
```suggestion
    server: Optional[str] = None,
```

app/repositories/maps.py
```diff
@@ -122,7 +122,7 @@ async def create(
 
 
 async def fetch_one(
```
cmyui, line 124: not sure about this for fetch_one, with the composite key, you could have overlapping beatmap ids between two servers

cmyui: interesting take actually

cmyui: you may want to check our current usage of `server="osu!"` throughout the codebase; i made them without this consideration

cmyui: @tsunyoku curious of your thoughts here (if any)

def750: > you may want to check our current usage of `server="osu!"` throughout the codebase; i made them without this consideration

Yeah I actually checked that, especially because it got quite irritating in our implementation of above and I needed to change it everywhere :kekw:.

I can include pr with changes of that  in beatmaps.py and other places if needed.

7ez: > you may want to check our current usage of `server="osu!"` throughout the codebase; i made them without this consideration

wasn't that for when eventually you would implement bss? atleast that's what I use it for

tsunyoku: @def750 @cmyui little bit confused about your guys' discussion.. we cannot have overlapping beatmap ids across servers because you can't possibly handle 2 maps with the same id in the osu! client.

i don't think this PR is an issue, but i don't really understand the comments about overlapping ids either. all this pr does, in my eyes, is allow you to fetch from both servers simultaneously (may be useful for direct if we ever have bss?)

cmyui:
thought about this a bit more and got some opinions from others - i'm going to make ids globally unique. if a server wants to re-submit a beatmap, they should do so under a new ID, and server will no longer be part of the primary key, but rather a regular column on the maps table.

app/repositories/maps.py
cmyui: okay yeah @tsunyoku i'm moving in that direction now #358

cmyui (approved):
+1 👍🏻 thanks!

PR #341, opened 2023-01-08
Adding GDPR functionality

minisbett:
This adds GDPR compliance for bancho.py to work as independently from other services. The gdpr.py in usecases handles collecting the data, converting them to the CSV format and putting them in a zip archive.

The /get_gdpr_data was added with features:
1. Accessing your own data by just specifying your API key
2. Accessing anyones data by specifying an API key of a user with developer privileges and the user id in the id parameter
3. Accessing your own data by using the GDPR code generated.

The !gdpr command generates a GDPR code that expires after 5 minutes. A link to the api endpoint with that code as the code parameter will be sent to the user.


Sukooo:
Why.

tsunyoku (changes requested):
i have a few concerns.

first off, would it not be better to export as json instead? this is far more readable to the average person and makes it easier to visualise data - bancho for example do this with their data.

second, keeping an internal cache for the access codes is bad UX in case of server restarts; it should go into the database. also the time window for a code should probably
be longer than 5 minutes, perhaps something like 30 days (took this example time from discord’s data email)

third, using oauth2 for this seems like trying to find a short coded way to implement auth. oauth2 shouldn’t really be used for an endpoint with temporary access without implementing proper oauth2 protocol (POST /oauth/token flow, etc.). so i suggest just having the “token” as an argument in the query, or implementing proper oauth.

fourth, i think you are providing *too much* data. GDPR is more about personally identifiable info, not literally everything. for example, providing any logs related to the user is not a good idea in my eyes nor is it necessary. for restrictions or other cases where logs are evidence against the user, these will only cause troubles sharing. internal things like that don’t need to be provided. same thing can be applied to the user’s hashes as it may help to bypass multiaccounts, however it is personally identifiable so i can let this one slip.

lastly, is there a particular reason you commented out the replay code? i don’t think saving this was particularly a bad idea.

minisbett:
> first off, would it not be better to export as json instead? this is far more readable to the average person and makes it easier to visualise data

I'd actually say the complete opposite. Especially with masses of scores it's far easier to load it into excel or google spreadsheet to have a clean list.

> second, keeping an internal cache for the access codes is bad UX in case of server restarts; it should go into the database. also the time window for a code should probably
be longer than 5 minutes, perhaps something like 30 days (took this example time from discord’s data email)

Not really needed, the data is only supposed to be accessable for 5 minutes anyways. As the GDPR command is more of a on-the-fly thing you just run and done, there shouldn't be any concerns for server restarts. The code is basically just a way of providing a file to download to the user directly in the in-game chat.

> fourth, i think you are providing too much data. GDPR is more about personally identifiable info, not literally everything. for example, providing any logs related to the user is not a good idea in my eyes nor is it necessary. for restrictions or other cases where logs are evidence against the user, these will only cause troubles sharing. internal things like that don’t need to be provided.

Legally they have to be provided. If servers would like to risk it they can comment out that table entry but generally all data associated with the user id has to be provided. And yes, even if that means that some internal logs are made public. That's how it works with all services. I got banned on hypixel once and when I requested GDPR there was various internal ban system data too. It's a legal requirement.

> lastly, is there a particular reason you commented out the replay code? i don’t think saving this was particularly a bad idea.

cmyui had some doubts whether this wouldn't just bloat everything, I think it belongs there but I followed his advice. That's why I made it optional for people to comment in, in case anyone messages them about their GDPR package not containing their replays.

> third, using oauth2 for this seems like trying to find a short coded way to implement auth. oauth2 shouldn’t really be used for an endpoint with temporary access without implementing proper oauth2 protocol (POST /oauth/token flow, etc.). so i suggest just having the “token” as an argument in the query, or implementing proper oauth.

cmyui helped me implementing the API key checks for the new webserver stuff since the old method was still using an old webserver (was really old). As he helped me with it, I'm assume he's fine with using that key to access the gdpr data.



Now that I think about it, maybe it would be better to instead of doing the code stuff at all just generate an api key if it does not exist and return a permanent link containing the api key instead?

tsunyoku:
> generally all data associated with the user id has to be provided

uhhh no. anything which is personally identifiable to. GDPR is tight, but not this tight because practicality has to exist too. if hypixel chose to share those logs great, but logs are not sharing personally identifiable and i will still hold the opinion they don’t need to be shared

take a read at https://gdpr-info.eu/issues/personal-data and likely other places too, just because it is attached to a user id doesn’t make it personal data if it doesn’t mean anything to the person themselves

minisbett:
> anything which is personally identifiable to

This is called PII (Personally Identifable Information) data. But PII data is not the only thing that has to be provided via GDPR request. There is also GPD (General Personal Data). This data includes everything that is *related* to an identifiable individual. This can be a user id, a username, some cookie token or whatever.

The term "personal data" is defined in Art. 4 (1). Personal data are any information which are **related** to an identified or identifiable natural person.

minisbett:
> personal data’ means any information relating to an identified or identifiable natural person (‘data subject’); an identifiable natural person is one who can be identified, directly or indirectly, in particular by reference to an identifier such as a name, an identification number, location data, an online identifier or to one or more factors specific to the physical, physiological, genetic, mental, economic, cultural or social identity of that natural person;

https://gdpr-info.eu/art-4-gdpr/

tsunyoku:
logs are not personal data. they do not include names, hardware, emails, etc.

yes, they may reference a user however they are not directly identifiable. if you read the given examples you quoted, you can clearly tell the type of information they are referring to by “personal information” and in comparison to logs to they are far more personal.

do you really, *actually*, think that GDPR is able to enforce this much? could you go to police and ask for all evidence towards your own past convictions via GDPR just because the evidence is linked to your case? no, because that evidence is only *related* to the user. it’s an extreme example, i know, but you can apply it way more realistically through e.g. being banned on bancho

i’ve requested my data from numerous places, and we can take bancho and discord as good examples since they are both not small and don’t have any legal related scandals. neither of these places have provided me with any data which is related to myself, but not directly personally identifiable. i don’t have bancho support logs, anticheat logs etc. related to my account. i also don’t have any internal logs from discord data either.

not providing these logs is not against gdpr, they are not personally identifiable

minisbett:
> logs are not personal data. they do not include names, hardware, emails, etc.

They include the users ID, therefore are data saved that are related to their identifyable individual.

> do you really, actually, think that GDPR is able to enforce this much

yep, it is!

> could you go to police and ask for all evidence towards your own past convictions via GDPR  just because the evidence is linked to your case?

yes, they're literally legally required to.

> no, because that evidence is only related to the user

yes, related is the keyword here. Related data has to be included. "‘personal data’ means any information **>relating<** to an identified or identifiable natural person"

> i don’t have bancho support logs, anticheat logs etc. related to my account.

I honestly think that's more of a peppy issue.

You might not be correctly wrong, iirc there are parts of he GDPR that allow to not disclose important data for public interest n shit. I'm not a lawyer, maybe it falls under these laws but what you've listed so far here does not allow any room for not disclosing the data.

> they are not personally identifiable

They are, they are attached to the user's id and therefore are considered personal data as per Art. 4(1)

minisbett:
Also keep in mind that the country-wide implementation of the GDPR slightly varies from country to country.

minisbett:
> In practice, this includes all data that can be assigned to a person in any way.

translated from the german translation of the GDPR website from above

minisbett:
That's exactly the reason why after personal data expires it has to either be completely deleted or anonymized, for example by wiping the user id from the sql database.

minisbett:
> Art. 15 GDPR enables you to obtain information from the person responsible as to whether they are actually processing data relating to your person and, if so, which ones. This includes all data and information related to your person (Art. 4 Para. 1 No. 1 GDPR) that is available to the person responsible. The right to information therefore not only refers to so-called master data such as name, address and date of birth, but also, for example, to the communication conducted with you and internal notes of the authority or the company, insofar as these contain personal data from you. Often there is content and meaning of information relating to a data subject, also only from the processing context (example: correspondence between the person responsible and the data subject). In this case, as a rule, the relevant documents must be returned in full (copies).

tsunyoku:
don’t really think either of us are qualified to be having these debates, we are basically just making our own interpretations of written documents at this point

unless someone who is actually qualified and processes GDPR has anything to weigh in, we should probably make this something more generic & kept behind server staff to generate on behalf of a user.

otherwise, i can see this easily being a disaster regardless of if we include logs or not because we don’t legally know GDPR to a T

NiceAesth:
I heavily disagree with the approach this PR has taken.

1. The provided functionality is at essence a data dumping tool. I do not agree with the naming used. (e.g.: see Discord: https://support.discord.com/hc/en-us/articles/360004027692)
2. Such functionality while it may tangentially have parts which could be part of this repo, this should not be accessible outside of a provided front-end, which brings me to my other point
3. Having access to data dumps on demand and instantaneously allocating server resources towards generating these dumps in the same application / service is a massive blunder. In an ideal scenario the API should at MOST put the user in a queue for data dump generation and should be notified when processing is completed within a reasonable timeframe (most websites provide a 30 day period).
4. My final issue with this implementation is the provided data. Something which shockingly has not been mentioned yet is the fact that the current data dumps include information regarding other users as well. A data dump of the sort should ONLY include information strictly related to the user processing the request, otherwise you are in breach of data laws (specifically, friend relationship dumps). This would only be ok in the case you do not view user IDs as identifiable information, which would contradict the need for this functionality in the first place.

The main takeaways from this PR to aid in future implementations would be to:
1. Use a more abstract name, European data laws are not unique, other regions have their own data laws with different names.
2. Do not generate data dumps on demand.
3. Anonymize information in dumps. Other users should not be identifiable via one's own dump.

While such a tool is a nice-to-have, this implementation has taken what I believe to be an incorrect approach.

minisbett:
> The provided functionality is at essence a data dumping tool

That's exactly what it's supposed to be lol.

> is should not be accessible outside of a provided front-end, which brings me to my other point

Yes it should be, I discussed it with cmyui. bancho.py should be able to operate on it's own without the need of a frontend.

> Having access to data dumps on demand and instantaneously allocating server resources towards generating these dumps in the same application / service is a massive blunder. In an ideal scenario the API should at MOST put the user in a queue for data dump generation and should be notified when processing is completed within a reasonable timeframe (most websites provide a 30 day period).

I would agree but we are talking about kilobytes of data that take a second to process. At most, this should be set on cooldown to prevent people from causing massive load on the application. 

> (most websites provide a 30 day period).

because that's the legal limit.

> A data dump of the sort should ONLY include information strictly related to the user processing the request

I agree with that, I'm unsure how this is handled as yes it also discloses information about someone else but that information of someone else is related to your own. Therefore I'm unsure. As far as I can see the only thing that should be removed would be the friendship table, we can do that.

NiceAesth:
was closing this intended? lol

PR #343, opened 2023-01-08
refactor: nullability at the top level for player's last np

cmyui:


cmyui:
a bit more verbose but more explicit/correct imo

app/commands.py
```diff
@@ -838,7 +838,7 @@ async def user(ctx: Context) -> Optional[str]:
         priv.name for priv in Privileges if p.priv & priv and bin(priv).count("1") == 1
     ][::-1]
 
-    if time.time() < p.last_np["timeout"]:
+    if p.last_np is not None and time.time() < p.last_np["timeout"]:
```
NiceAesth, line 841: Not sure I understand what this is trying to achieve at first glance

cmyui: basically, rather than setting last_np as an dict with empty params by default (like timeout:0), we set nullable at the top level, last_np: Optional[LastNp] if the player has not nped a map yet

NiceAesth: yeah, i got that part but i wasn't sure why this particular check was necessary. changes look fine to me either way

NiceAesth (approved):

PR #346, opened 2023-01-08
fix: calculate aggregate pp & acc using all best scores

### LINKED ISSUE #299: [BUG] Overall weighted acc calculation returning wrong values
Since some time we've been experiencing weird issues with weighted acc calculation, we've had some users that were aiming for profile acc of 100%. But after some time acc went down even though they had only SS ranks.
Here's one of our users, his scores and output of acc calculation
https://pastebin.com/yYEG1H1Y

I used bancho.py's acc calculation method and fetched scores where status=2 in mode=4

cmyui:
previously only used top 100, which does not match behaviour of official servers

closes #299


NiceAesth (approved):
Changes look alright to me, I assume the performance impact has already been considered and deemed a non-issue

cmyui:
needs review from @tsunyoku as well as testing before merge

cmyui:
@NiceAesth yeah a recalculation of aggregate pp (& acc) will be required. we don't actually have a script to do acc.. perhaps this should be considered before a merge

tsunyoku:
> @NiceAesth yeah a recalculation of aggregate pp (& acc) will be required. we don't actually have a script to do acc.. perhaps this should be considered before a merge

this is already part of the recent pp recalc script. we could perhaps allow them to specify they only want to recalc stats and not scores?

tsunyoku:
infact, the script will need updating to reflect this change

tsunyoku:
also the !recalc cmd if that still exists & calculates user stats too, forgot entirely how that cmd works

i think that covers all possible places these calculations are ran?

cmyui:
> this is already part of the recent pp recalc script.

Ahh right I forgot it included acc since the ordering of scores can change from the recalc.

> we could perhaps allow them to specify they only want to recalc stats and not scores?

Yes let's do this

> also the !recalc cmd if that still exists

I've removed it

tsunyoku:
going to add to this pr to allow stats-only recalc in the recalc tool, and will go ahead and merge. will also make a mention in #breaking-changes

NiceAesth:
@tsunyoku Added blocked tag on this until the change is documented. Let me know when you are ready to revisit this PR, since I intend to create a page for breaking changes on the wiki.

tsunyoku:
> @tsunyoku Added blocked tag on this until the change is documented. Let me know when you are ready to revisit this PR, since I intend to create a page for breaking changes on the wiki.

all good on my end, the recalc script has been adjusted to allow only recalculating scores and/or stats so aslong as the change is documented there is an easy method for servers to recalculate their users

NiceAesth:
Oops, this should have been merged a while ago since wiki reflects this now.

tools/recalc.py
```diff
@@ -212,9 +212,23 @@ async def main(argv: Optional[Sequence[str]] = None) -> int:
...
+        description="Recalculate scores",
+        action="store_true",
+        default=True,
+    )
+    parser.add_argument(
+        "--stats",
+        description="Recalculate stats",
+        action="store_true",
+        default=True,
+    )
```
cmyui, line 231: In hindsight.. is it even possible to disable this now? I think with `store_true` this is strictly only able to be set to True..

Perhaps we should flip these to `--no-stats-recalc` `--no-scores-recalc` with a default of False

tsunyoku: lol stupid oversight on my end i think
i hate argparse

PR #348, opened 2023-01-09
feat: add `/calculate_pp` API endpoint; extend performance usecase data

minisbett:
Added the /calculate_pp endpoint. It takes the api key in the Authorization header.

In the query parameters, beatmap_id is the only required parameter. mods, mode, misses etc. all have a default value of 0, combo has the maximum combo of the beatmap as the default.

The return type of the endpoint is a single JSON object if the n100 and n50 parameters were used. If the acc parameter is used instead, the user can specify multiple accuracies in the format "99.1,98.38,88.23". In that case, a list of JSON objects is returned.
While changing return types on API endpoints are usually bad, we can do it here because the developer explicitly either requests a singlg object via n100 and n50 or multiple via the acc parameter. Therefore the developer can expect what the return type is.

The extra n100 and n50 is necessary as the pp system has a 50 hit penalty. If you specify accuracies, rosu-pp automatically assumes the best-case-scenario for the 50 count. This is usually desired when requesting a list of accuracies, e.g. for visualizing the pp values for common accuracies (100, 99, 98, 95, ...)

I think adding the possibility of specfying multiple n100 and n50 pairs would bloat the endpoint with parsing and does not really have much real world applications. And since the performance usecase does not optimize for calculating multiple scores anyways (e.g. by caching star ratings) the user could just send multiple requests to the endpoint if more than one n100/n50 pair is necessary.

I also reworked the return of the performance usecase and adapted the code that uses the calculate_performances() method to it. It now returns all performance and difficulty-related values instead of just the pp and star rating.


app/api/domains/api.py
```diff
@@ -109,6 +116,77 @@ def format_map_basic(m: Beatmap) -> dict[str, object]:
...
+async def api_calculate_pp(
+    token: HTTPCredentials = Depends(oauth2_scheme),
+    beatmap_id: int = Query(None, alias="id", min=0, max=2_147_483_647),
+    n100: int = Query(0, max=2_147_483_647),
+    n50: int = Query(0, max=2_147_483_647),
+    misses: int = Query(0, max=2_147_483_647),
+    mods: int = Query(0, min=0, max=2_147_483_647),
+    mode: int = Query(0, min=0, max=11),
+    combo: int = Query(None, max=2_147_483_647),
+    acclist: Optional[str] = Query(None, alias="acc"),
```
cmyui, line 129: hey, better to use industry standard here for passing multiple query parameters,
`?accuracy=12.34&accuracy=32.65&accuracy=43.76`

this can be accomplished by using `list[str]` as the type, fastapi will handle the magic for you

cmyui: actually, you can probably use `list[float]`

minisbett: oh alright

app/api/domains/api.py
```diff
@@ -126,11 +126,11 @@ async def api_calculate_pp(
     mods: int = Query(0, min=0, max=2_147_483_647),
     mode: int = Query(0, min=0, max=11),
     combo: int = Query(None, max=2_147_483_647),
-    acclist: Optional[str] = Query(None, alias="acc")
+    acclist: list[float] = Query(None, alias="acc")
```
cmyui: @infernalfire72 hey do you remember off the top of your head if this default `None` is correct? i always forget how this works in fastapi

minisbett: I replaced it now anyways, will do another push soon

tsunyoku: double checked with fastapi docs: https://fastapi.tiangolo.com/tutorial/query-params-str-validations/

u should use `Query([], alias="acc")`

cmyui: That kinda makes me wonder if they pass the same list instance all the way down the stack, in that case it could be susceptible to the same issues as mutable parameter defaults LOL

tsunyoku: yeah it would be if true, doesn’t matter in our case tho.. we’re not mutating the list, surely?

cmyui: I'll test this later, I need to know now LOL.

If it does behave that way, best to use `default_factory` as to not promote an antipattern.

NiceAesth:
@osuAkatsuki/bancho-py-devs has anyone looked over this PR since it has been updated yet?

cmyui:
seems fine for v1

cmyui (approved):

PR #349, opened 2023-01-09
feat: rename `get_apikey` to `apikey`; provide API key via browser

minisbett:
This is one of the things that always annoyed me, I renamed !get_apikey to just !apikey since the underscore feels out of place.
I also changed doing /savelog to a http://<api key> hyperlink such that you can just click and copy the text in the browser bar. Feels faster for me than typing /savelog, clicking the notification and marking the text in notepad precisely.


app/commands.py
```diff
@@ -571,12 +571,7 @@ async def get_apikey(ctx: Context) -> Optional[str]:
     await players_repo.update(ctx.player.id, api_key=ctx.player.api_key)
     app.state.sessions.api_keys[ctx.player.api_key] = ctx.player.id
 
-    ctx.player.enqueue(
-        app.packets.notification(
-            "Type /savelog and click the popup for an easy way to copy this.",
-        ),
-    )
-    return f"Your API key is now: {ctx.player.api_key}"
+    return f"API key generated. Click (here)[http://{ctx.player.api_key}] to view your new API key."
```
cmyui, line 574:
```suggestion
    return f"API key generated. Copy your api key from (this url)[http://{ctx.player.api_key}]."
```

cmyui (approved):

PR #352, opened 2023-01-09
feat: support all judgements in pp calculation

def750:
Changed calculator to accept all judgements as they have critical meaning in calculation especially in mania, or 50s in std, which weren't included This is also done on score submission in @ https://github.com/osuAkatsuki/bancho.py/blob/master/app/objects/score.py#L333-L345

Also made SQL Query for scores more readable


def750:
> is this code tested?

Now yes :trollface:

cmyui:
hey @def750, is this good to go?

also, have you been running this code in production on your server?

cmyui (approved):

PR #355, opened 2023-01-09
chore: add mitmproxy and mypy as dev dependendencies 

NiceAesth:


requirements.txt
```diff
@@ -13,7 +13,7 @@ charset-normalizer==2.1.1 ; python_full_version >= '3.6.0'
 click==8.1.3 ; python_version >= '3.7'
 databases[mysql]==0.7.0
 datadog==0.44.0
-fastapi==0.88.0
+fastapi==0.89.0
```
cmyui, line 16: this includes a breaking change for return type annotations on api endpoints, we should make sure no endpoints have any return annotations 

https://github.com/tiangolo/fastapi/releases/tag/0.89.0

NiceAesth: checked everything that had the `@router` decorator and i can't seem to find any that have annotations

Pipfile
```diff
@@ -30,6 +30,7 @@ pytest = "*"
 pre-commit = "*"
 black = "*"
 reorder-python-imports = "*"
+mitmproxy = "*"
```
cmyui, line 33: no mypy?

NiceAesth: huh?? i guess my changes got discarded, fixing

NiceAesth: fixed, looking like a squash and merge angle

cmyui (approved):

PR #356, opened 2023-01-10
refactor: simplify `!with` command argument parsing

minisbett:
I simplified the parsing, as well as adding some more "implicit" parameters to make stuff more consistent.
I also added support for mania (katu, geki)
In summary, you can now do stuff such as

!with 5x100 2x50
!with 99% 3m
!with 4x100 9m 388x
!with 97.59% 2m 473x +HDDT
!with 5xkatu 3xgeki

Dependent on #526


minisbett:
should be good to merge now

minisbett:
> setattr is woozy, generally not a fan of this pr being in favour of the current logic. open to opinions from other reviewers though

It's a fair point, although I think it's fair to use in such a use-case of dynamically parsing user input like this. The logic is much simpler, the command itself is easier to use imo and it also adds missing support.

minisbett:
replaced by #640

PR #357, opened 2023-01-10
Handle / suffix on mirror url

minisbett:
Removing a "/" at the end of mirror urls on environment variable load since this has been a really frequent issue developers faced.


cmyui:
nah this is standard to not include trailing /, i've had this debate before but don't care to handle user input to this degree - there are many other places we could start mini debates like this with user input, my opinion is to not sanitize it at all

maybe in the future if we want to go VERY user friendly

PR #358, opened 2023-01-10
Globally unique map ids

cmyui:


cmyui (commented):
@def750 @alowave223 this might be quite expensive on larger servers, so i've added locks to make sure guweb (or alike) doesn't write to it between the queries

i believe this should work; pls let me know if anything goes weird

PR #360, opened 2023-01-10
Add reason to unsilence command

minisbett:
One of the things I added on cs0su for convenience, would also make it consistent as there are also reasons for unrestrictions

### DIFF
--- a/app/commands.py
+++ b/app/commands.py
@@ -803,8 +803,8 @@ async def silence(ctx: Context) -> Optional[str]:
 @command(Privileges.MODERATOR, hidden=True)
 async def unsilence(ctx: Context) -> Optional[str]:
     """Unsilence a specified player."""
-    if len(ctx.args) != 1:
-        return "Invalid syntax: !unsilence <name>"
+    if len(ctx.args) < 2:
+        return "Invalid syntax: !unsilence <name> <reason>"
 
     if not (t := await app.state.sessions.players.from_cache_or_sql(name=ctx.args[0])):
         return f'"{ctx.args[0]}" not found.'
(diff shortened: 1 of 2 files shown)


cmyui:
@osuAkatsuki/bancho-py-devs think there's value to this?

minisbett:
> @osuAkatsuki/bancho-py-devs think there's value to this?

Idk maybe there's a usage, pretty much the same there'd be for restrictions. Would be weird if it's on !unrestrict but not !unsilence

app/commands.py
```diff
@@ -826,8 +826,8 @@ async def unsilence(ctx: Context) -> Optional[str]:
     if target.priv & Privileges.STAFF and not ctx.player.priv & Privileges.DEVELOPER:
         return "Only developers can manage staff members."
 
-    await target.unsilence(ctx.player)
-    return f"{target} was unsilenced."
+    await t.unsilence(ctx.player, " ".join(ctx.args[1:]))
```
NiceAesth, line 829: huh?

minisbett: oh I fucked up the merge conflict 1 sec

cmyui: we should at a minimum
```suggestion
    await t.unsilence(ctx.player, reason=" ".join(ctx.args[1:]))
```

minisbett: why that

minisbett: unrestrict doesn't have it either

cmyui:
> Idk maybe there's a usage, pretty much the same there'd be for unrestrictions. Would be weird if it's on !unrestrict but not !unsilence

i think there was already limited value to unrestrict message (generally was used to say "appealed" on akatsuki, but was already getting spammed with like "a" and "xd" and stuff), but for unsilence.. really seems like we're picking at straws

app/commands.py
NiceAesth: speaking of unrestrict, i am not a fan of having the join inline. would rather you put in in a variable

minisbett: fixed

NiceAesth:
well, i am fine with the impl, but i am kinda indifferent towards whether or not this is needed. would need other opinions

cmyui (approved):

PR #361, opened 2023-01-10
 Fix not being able to pick converts maps in multiplayer 

7mochi:
### Expected behaviour
Pick in a mp match a std map and play it in another gamemode (catch/taiko/mania) without any problem

### Observed behaviour
Pick a std map, choose another gamemode and fallback to std.

### Videos
- Bug

https://user-images.githubusercontent.com/33896771/211444216-ba02ce9b-cb3b-401c-8e16-eeddb68fac2d.mp4

- With this PR

https://user-images.githubusercontent.com/33896771/211444288-9b2fc661-9126-45fb-870d-947ba0cb657b.mp4

### DIFF
--- a/app/api/domains/cho.py
+++ b/app/api/domains/cho.py
@@ -1534,7 +1534,7 @@ class MatchChangeSettings(BasePacket):
                 player.match.map_id = bmap.id
                 player.match.map_md5 = bmap.md5
                 player.match.map_name = bmap.full_name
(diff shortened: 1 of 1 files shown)


cmyui (approved):
nice, i thought this would be more complex tbh. didn't know osu client blocked you from picking other modes on mode-specific maps.

maaaybe we should add some validation here? would suck if you can crash clients or something, but can come later

PR #365, opened 2023-01-10
separate api v1 from v2; GET v2/players endpoints

cmyui:


app/api/v1/__init__.py
```diff
@@ -0,0 +1,11 @@
+# type: ignore
+from __future__ import annotations
+
+from fastapi import APIRouter
+
+from .api import router
+
+
+apiv1_router = APIRouter(tags=["API-v1"], prefix="/v1")
```
cmyui, line 9: i'm not much of a fan of prefixes at levels like this, i think one of the benefits of decorators in python is being able to put the endpoint meta info right at the function itself

this is sort of like the old pattern, where you'd make a bunch of api functions, then include them all at this (separate) router level

instead, let's put everything at the level of the function, including the full path to the api endpoint

app/api/v2/common/responses.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class ErrorResponse(GenericModel, Generic[T]):
+    status: Literal["error"]
+    error: T
+    message: str
+
+
+def failure(
+    # TODO: error code
```
cmyui, line 42: probably should do this before merge

app/api/v1/__init__.py
cmyui: will do this later; want to merge this quick so you can continue work

app/api/v2/common/responses.py
cmyui: for later, actually

PR #366, opened 2023-01-10
refactor: api-v2 authorization

alowave223:
Here, again, referring to the Akatsuki-v2 I've implemented the basics of authorization logic. There are many things TODO but I think it's better to make small PR's so we can keep a track of what we are doing.


app/api/v2/models/sessions.py
```diff
@@ -0,0 +1,41 @@
...
+from typing import Union
+from uuid import UUID
+
+from . import BaseModel
+
+# input models
+
+
+class SessionUpdate(BaseModel):
+    expires_at: Union[datetime, None]
```
cmyui, line 15: prefer optional syntax for 3.9

app/api/v2/sessions.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+@router.post("/sessions")
+async def create_session(args: LoginForm) -> Success[Session]:
+    data = await sessions.authorize(username=args.username, password=args.password)
+    if data is None:
+        return responses.failure(message="Failed to create session.")
+
+    response = Session.from_mapping(data)
+    return responses.success(response)
```
cmyui, line 30: i think we want to implement oauth with client id and secret rather than using name & password like this

see osu api v2 for example

app/api/v2/sessions.py
```diff
@@ -0,0 +1,73 @@
...
+
+@router.patch("/sessions/self")
+async def patch_session(
+    args: SessionUpdate,
+    token: HTTPCredentials = Depends(oauth2_scheme),
+) -> Success[Session]:
+    session = await sessions_repo.fetch_one(uuid.UUID(token.credentials))
+    if session is None:
+        return responses.failure(message="Couldn't find session.")
+
```
cmyui, line 53: should be able to just use partial_update alone
```suggestion
```

app/api/v2/sessions.py
```diff
@@ -0,0 +1,73 @@
...
+    return responses.success(data)
+
+
+@router.delete("/sessions")
+async def delete_session(
+    token: HTTPCredentials = Depends(oauth2_scheme),
+) -> Success[Session]:
+    data = await sessions.deauthorize(uuid.UUID(token.credentials))
+    if data is None:
+        return responses.failure(message="Failed to create session.")
```
cmyui, line 70:
```suggestion
        return responses.failure(message="Failed to deletesession.")
```

app/api/v2/sessions.py
```diff
@@ -0,0 +1,73 @@
...
+        uuid.UUID(token.credentials),
+        expires_at=args.expires_at,
+    )
+    if data is None:
+        return responses.failure(message="Couldn't update session.")
+
+    return responses.success(data)
+
+
+@router.delete("/sessions")
```
cmyui, line 64: tbh i might remove the /self things later & replace with /{session_id} but it's fine for now
```suggestion
@router.delete("/sessions/self")
```

app/repositories/sessions.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+async def fetch_one(session_id: UUID) -> Union[Mapping[str, Any], None]:
+    session = await app.state.services.redis.get(create_session_key(session_id))
+    if session is None:
+        return None
+    return json.loads(session)
+
+
+async def partial_update(session_id: UUID, **kwargs: Any) -> Mapping[str, Any] | None:
```
cmyui, line 45: 3.10 syntax

cmyui (changes requested):
Will need changes, the usecases abstraction was something i was looking to avoid until later (like we did in players apis)

also will need to switch to proper oauth

@NiceAesth u have any free cycles to work on this with me any time soon?

cmyui:
@alowave223 i think it's best if you continue with non-authed endpoints like GETs for other resources - there are plenty to do outside of this for now, notably;

players [DONE]
maps
scores
stats
clans
channels
tourney_pools
tourney_pool_maps

NiceAesth:
@cmyui I am a bit busier than usual these days, but I can try something either thursday or in the weekends. (I have exams upcoming so this month so yeah)

alowave223:
Will be re-done

PR #367, opened 2023-01-10
refactor: move geolocation system (maxmind) to nginx layer

### LINKED ISSUE #267: Move geolocation system (maxmind) to nginx layer
This is something that's better handled by nginx - it provides a [built-in module for this](http://nginx.org/en/docs/http/ngx_http_geoip_module.html).

This lets us cut the database file, code, and project dependency out of our codebase.

7mochi:
Closes #267 
- [x] Check if default value is present on headers in fetch_nginx method to see if maxminddb is enabled (remove .env toggle)
- [x] Remove geoip2 and maxminddb dependencies and regenerate requeriments

Tested on:
- [x] localhost without nginx module
- [x] localhost with nginx module
- [x] remote without nginx module
- [x] remote with nginx module
- [x] remote with cloudflare


cmyui:
awww yeaaaa

NiceAesth:
shouldn't dependencies also be updated?

7mochi:
> shouldn't dependencies also be updated?

True, also the module must be loaded in nginx.conf

7mochi:
I think it's ready now

tsunyoku (approved):

NiceAesth:
is maxminddb still required in pipfile?

NiceAesth:
@7mochi

app/api/domains/cho.py
```diff
@@ -452,6 +458,7 @@ def parse_login_data(data: bytes) -> LoginData:
 
 async def login(
     body: bytes,
+    headers: Mapping[str, str],
```
cmyui, line 461:
```suggestion
    headers: Mapping[str, str],
    body: bytes,
```

app/api/domains/cho.py
```diff
@@ -111,7 +112,12 @@ async def bancho_handler(
     if osu_token is None:
         # the client is performing a login
         async with app.state.services.database.connection() as db_conn:
-            login_data = await login(await request.body(), ip, db_conn)
+            login_data = await login(
+                await request.body(),
+                request.headers,
```
cmyui, line 117:
```suggestion
                request.headers,
                await request.body(),
```

ext/nginx.conf
```diff
@@ -25,6 +36,9 @@ server {
 	location / {
 		proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
 		proxy_set_header X-Real-IP  $remote_addr;
+		proxy_set_header X-Country-Code $geoip2_data_country_code;
+		proxy_set_header X-Latitude $geoip2_data_latitude;
+		proxy_set_header X-Longitude $geoip2_data_longitude;
```
cmyui, line 41: should this be commented out by default? (will this throw if the module is not enabled?)

NiceAesth: would bpy not crap out either way if it wasn't enabled?

cmyui: it has an api-based backup approach if it is disabled. (we need to ensure that's actually working)

cmyui: see https://github.com/osuAkatsuki/bancho.py/pull/367/files#diff-4534a559c52f8fe8ea43b1b8d0649adef038fc8530855d8c7234c7ad5372cac8R1954

NiceAesth: also i have a suggestion for this then:

1. set default value for those variables
2. check if default value is present on those headers in fetch_nginx to see if maxminddb enabled

NiceAesth:
```nginx
set $geoip2_data_country_code 'default';
...
```

NiceAesth (changes requested):
see previous comments

would require updating `Pipfile` and rebuilding dependencies at the very least

ext/nginx.conf
7mochi: Default values are here https://github.com/osuAkatsuki/bancho.py/pull/367/files#diff-74e6757613db9ff8909a81b274ee6a836e1429d23734a01820bd0a00194315afR18-R20
I agree with the second point

7mochi:
> is maxminddb still required in pipfile?



> is maxminddb still required in pipfile?

No longer required

ext/nginx.conf
NiceAesth: oh, hadn't noticed defaults were set

7mochi: > should this be commented out by default? (will this throw if the module is not enabled?)

if the module isn't loaded or installed nginx will throw this error
image

app/api/domains/cho.py
```diff
@@ -719,14 +719,14 @@ async def login(
...
     if not ip.is_private:
-        if app.settings.NGINX_GEOIP:
-            # good, dev has downloaded a geoloc db from maxmind,
-            # so we can do a local db lookup. (typically ~1-5ms)
-            # https://www.maxmind.com/en/home
-            geoloc = app.state.services.fetch_geoloc_nginx(ip, headers)
-        else:
-            # bad, we must do an external db lookup using
-            # a public api. (depends, `ping ip-api.com`)
+        # good, dev has downloaded a geoloc db from maxmind,
```
NiceAesth, line 722: these comments don't really mean anything anymore

app/state/services.py
```diff
@@ -137,25 +131,26 @@ def get_ip(self, headers: Mapping[str, str]) -> IPAddress:
...
-        },
-    }
+def fetch_geoloc_nginx(
+    ip: IPAddress,
+    headers: Mapping[str, str],
+) -> Optional[Geolocation]:
+    """Fetch geolocation data based on ip (using nginx headers)."""
+    country_code = headers.get("X-Country-Code")
+
+    if country_code is not None:
```
NiceAesth, line 141: it is a lot cleaner if you do failure checks and return instead of putting the entire function in an if statement.

```py
if country_code is None:
    return
```

app/api/domains/osu.py
```diff
@@ -1951,14 +1951,9 @@ async def register_account(
...
-                    geoloc = app.state.services.fetch_geoloc_db(ip)
-                else:
-                    # worst case, we must do an external db lookup
-                    # using a public api. (depends, `ping ip-api.com`)
+                geoloc = app.state.services.fetch_geoloc_nginx(ip, request.headers)
+
+                if geoloc is None:
                     geoloc = await app.state.services.fetch_geoloc_web(ip)
 
                 if geoloc is not None:
```
NiceAesth, line 1959: would this ever be reached even? localhost + no db or what? i would like some more testing on this PR

```py
if geoloc is not None:
```

also i think while we are refactoring this we could have this a method that parses headers

1. cloudflare first
2. nginx headers second
3. XX as fallback

and move logic outside from here as it could potentially be used elsewhere as well

app/api/domains/cho.py
```diff
@@ -712,14 +719,9 @@ async def login(
...
     if not ip.is_private:
-        if app.state.services.geoloc_db is not None:
-            # good, dev has downloaded a geoloc db from maxmind,
-            # so we can do a local db lookup. (typically ~1-5ms)
-            # https://www.maxmind.com/en/home
-            geoloc = app.state.services.fetch_geoloc_db(ip)
-        else:
-            # bad, we must do an external db lookup using
-            # a public api. (depends, `ping ip-api.com`)
+        geoloc = app.state.services.fetch_geoloc_nginx(ip, headers)
```
NiceAesth, line 722: why does login not check for cloudflare headers? i remember people mentioning they had to do some weird configs on cloudflare for it to work. i think this might be the reason why. also if we decouple the IP lookup strategy logic from ``domains/osu.py`` it could be reused here.

app/state/services.py
```diff
@@ -137,25 +131,26 @@ def get_ip(self, headers: Mapping[str, str]) -> IPAddress:
...
+        country_code = country_code.lower()
+        latitude = headers.get("X-Latitude")
```
NiceAesth, line 143: this will fail if the headers somehow aren't present if we are using `get()` as it will return `None` which cannot be cast to float. unsure if we even want to allow this:

1. either use `["X-Latitude"]`
2. or do `.get("X-Latitude", 0)` or whatever default you wish to use

NiceAesth (dismissed):

NiceAesth:
@cmyui would you be available to try the tests i requested? (look original post). could also be tried after the changes i suggested

app/api/domains/osu.py
cmyui: it can be reached if web api is in use & is down

cmyui: hmm i think i like the 1/2/3 idea

NiceAesth:
should probably add instructions on how to enable cloudflare geolocation data btw ( https://support.cloudflare.com/hc/en-us/articles/200168236-Configuring-IP-geolocation#12345680 )

app/state/services.py
```diff
@@ -137,23 +131,66 @@ def get_ip(self, headers: Mapping[str, str]) -> IPAddress:
...
+    headers: Mapping[str, str],
+) -> Optional[Geolocation]:
+    """Fetch geolocation data based on ip (using cloudflare headers)."""
+    country_code = headers.get("CF-IPCountry", "xx")
 
-    if res.country.iso_code is not None:
-        acronym = res.country.iso_code.lower()
-    else:
-        acronym = "XX"
+    if country_code is None:
```
NiceAesth, line 157: country_code will never be none. check if CF-IPCountry, CF-IPLatitude or CF-IPLongitude are missing and return. otherwise go on with the code (drop the gets as well

app/state/services.py
```diff
@@ -137,23 +131,66 @@ def get_ip(self, headers: Mapping[str, str]) -> IPAddress:
...
+        },
+    }
+
+
+def fetch_geoloc_nginx(
+    ip: IPAddress,
+    headers: Mapping[str, str],
+) -> Optional[Geolocation]:
+    """Fetch geolocation data based on ip (using nginx headers)."""
+    country_code = headers.get("X-Country-Code", "xx")
```
NiceAesth, line 179: same thing as cf comment

app/state/services.py
```diff
@@ -137,23 +131,66 @@ def get_ip(self, headers: Mapping[str, str]) -> IPAddress:
...
+    ):
+        return
+
+    country_code = headers["X-Country-Code"].lower()
+    latitude = headers["X-Latitude"]
+    longitude = headers["X-Longitude"]
+
+    return {
+        "latitude": float(latitude),
+        "longitude": float(longitude),
```
NiceAesth, line 190:
```suggestion
        "longitude": longitude,
```

app/state/services.py
```diff
@@ -137,23 +131,66 @@ def get_ip(self, headers: Mapping[str, str]) -> IPAddress:
...
+        key in headers for key in ("X-Country-Code", "X-Latitude", "X-Longitude")
+    ):
+        return
+
+    country_code = headers["X-Country-Code"].lower()
+    latitude = headers["X-Latitude"]
+    longitude = headers["X-Longitude"]
+
+    return {
+        "latitude": float(latitude),
```
NiceAesth, line 189:
```suggestion
        "latitude": latitude,
```

app/state/services.py
```diff
@@ -137,23 +131,66 @@ def get_ip(self, headers: Mapping[str, str]) -> IPAddress:
...
+) -> Optional[Geolocation]:
+    """Fetch geolocation data based on ip (using nginx headers)."""
+    if not all(
+        key in headers for key in ("X-Country-Code", "X-Latitude", "X-Longitude")
+    ):
+        return
+
+    country_code = headers["X-Country-Code"].lower()
+    latitude = headers["X-Latitude"]
+    longitude = headers["X-Longitude"]
```
NiceAesth, line 186:
```suggestion
    longitude = float(headers["X-Longitude"])
```

app/state/services.py
```diff
@@ -137,23 +131,66 @@ def get_ip(self, headers: Mapping[str, str]) -> IPAddress:
...
+    headers: Mapping[str, str],
+) -> Optional[Geolocation]:
+    """Fetch geolocation data based on ip (using nginx headers)."""
+    if not all(
+        key in headers for key in ("X-Country-Code", "X-Latitude", "X-Longitude")
+    ):
+        return
+
+    country_code = headers["X-Country-Code"].lower()
+    latitude = headers["X-Latitude"]
```
NiceAesth, line 185:
```suggestion
    latitude = float(headers["X-Latitude"])
```

app/state/services.py
```diff
@@ -137,23 +131,66 @@ def get_ip(self, headers: Mapping[str, str]) -> IPAddress:
...
+
+    country_code = headers["CF-IPCountry"].lower()
+    latitude = headers["CF-IPLatitude"]
+    longitude = headers["CF-IPLongitude"]
 
     return {
-        "latitude": res.location.latitude or 0.0,
-        "longitude": res.location.longitude or 0.0,
+        "latitude": float(latitude),
+        "longitude": float(longitude),
```
NiceAesth, line 166:
```suggestion
        "longitude": longitude,
```

app/state/services.py
```diff
@@ -137,23 +131,66 @@ def get_ip(self, headers: Mapping[str, str]) -> IPAddress:
...
+        return
+
+    country_code = headers["CF-IPCountry"].lower()
+    latitude = headers["CF-IPLatitude"]
+    longitude = headers["CF-IPLongitude"]
 
     return {
-        "latitude": res.location.latitude or 0.0,
-        "longitude": res.location.longitude or 0.0,
+        "latitude": float(latitude),
```
NiceAesth, line 165:
```suggestion
        "latitude": latitude,
```

app/state/services.py
```diff
@@ -137,23 +131,66 @@ def get_ip(self, headers: Mapping[str, str]) -> IPAddress:
...
+) -> Optional[Geolocation]:
+    """Fetch geolocation data based on ip (using cloudflare headers)."""
+    if not all(
+        key in headers for key in ("CF-IPCountry", "CF-IPLatitude", "CF-IPLongitude")
+    ):
+        return
+
+    country_code = headers["CF-IPCountry"].lower()
+    latitude = headers["CF-IPLatitude"]
+    longitude = headers["CF-IPLongitude"]
```
NiceAesth, line 162:
```suggestion
    longitude = float(headers["CF-IPLongitude"])
```

app/state/services.py
```diff
@@ -137,23 +131,66 @@ def get_ip(self, headers: Mapping[str, str]) -> IPAddress:
...
+    headers: Mapping[str, str],
+) -> Optional[Geolocation]:
+    """Fetch geolocation data based on ip (using cloudflare headers)."""
+    if not all(
+        key in headers for key in ("CF-IPCountry", "CF-IPLatitude", "CF-IPLongitude")
+    ):
+        return
+
+    country_code = headers["CF-IPCountry"].lower()
+    latitude = headers["CF-IPLatitude"]
```
NiceAesth, line 161:
```suggestion
    latitude = float(headers["CF-IPLatitude"])
```

NiceAesth (changes requested):

NiceAesth (approved):
changes look fine to me

7mochi:
> @cmyui would you be available to try the tests i requested? (look original post). could also be tried after the changes i suggested

i have just tested all cases and they work correctly

i think it's ready to merge

NiceAesth:
up to cmyui, I don't have merge perms but it looks alright to me

README.md
```diff
@@ -201,6 +205,12 @@ and you should see something along the lines of:
 
 ![tada](https://cdn.discordapp.com/attachments/616400094408736779/993705619498467369/ld-iZXysVXqwhM8.png)
 
+
+## enabling cloudflare geolocation data
```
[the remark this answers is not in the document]
7mochi: fixed

NiceAesth (approved):
ok

7mochi:
👀

cmyui (approved):

cmyui:
+1 👍🏻

NiceAesth:
LGTM :shipit:

app/state/services.py
```diff
@@ -137,30 +131,76 @@ def get_ip(self, headers: Mapping[str, str]) -> IPAddress:
...
     }
 
 
 async def fetch_geoloc_web(ip: IPAddress) -> Optional[Geolocation]:
     """Fetch geolocation data based on ip (using ip-api)."""
-    url = f"http://ip-api.com/line/{ip}"
+    if not ip.is_private:
+        url = f"http://ip-api.com/line/{ip}"
+    else:
+        url = "http://ip-api.com/line/"
```
cmyui, line 203: Noticed this too late - but we should actually be using `if not ip.is_loopback` here -- not `ip.is_private`

cmyui: Resolved here: https://github.com/osuAkatsuki/bancho.py/commit/112012a4a39c6cbb39f5a19a7523b084d70cc3cf

PR #368, opened 2023-01-11
refactor: generalize beatmap mirror code

minisbett:
Removed all mirror related code, which was partly old code that does not even apply anymore anyways. I talked with kotrik about a flaw in chimu's cheesegull that he will fix.

I also split up the search and download endpoint as it might be beneficial for some to use different mirrors for search and download, e.g. if you want mino/kitsu search speeds but use chimus big beatmap cache.

The code is tested with uk.catboy.best and kitsu.moe, chimu.moe can be tested after kotrik renamed "SetId" to "SetID".

NiceAesth and I agree that mirror related could should not be in bpy and that it's every mirrors own responsibility to provide proper cheesegull support.

I also set Mino to the default mirror, which can be discussed on whether we want that in bpy or not.


app/api/domains/osu.py
```diff
@@ -1784,14 +1747,11 @@ async def get_osz(
     if no_video:
         map_set_id = map_set_id[:-1]
 
-    if USING_CHIMU:
-        query_str = f"download/{map_set_id}?n={int(not no_video)}"
-    else:
-        query_str = f"d/{map_set_id}"
+    query_str = f"{map_set_id}?n={int(not no_video)}"
```
minisbett, line 1752: This can be generalized, no matter whether mirrors support it or not since otherwise this query parameter should be ignored (and is ignored on all mirrors I tested)

app/api/domains/osu.py
```diff
@@ -518,11 +486,8 @@ async def osuSearchHandler(
         if bmap["ChildrenBeatmaps"] is None:
             continue
 
-        if USING_CHIMU:
-            bmap["HasVideo"] = int(bmap["HasVideo"])
-        else:
-            # cheesegull doesn't support vids
-            bmap["HasVideo"] = "0"
+        # some mirrors use a true/false instead of 0 or 1
+        bmap["HasVideo"] = int(bmap["HasVideo"])
```
minisbett, line 490: Instead of checking whether it's chimu, do this for every mirror as chimu is not the only mirror providing the HasVideo property as a true/false value. Kitsu for example does that too

app/api/domains/osu.py
```diff
@@ -484,31 +469,14 @@ async def osuSearchHandler(
...
-    if USING_NASUYA:
-        # nasuya can serialize to direct for us
-        params["osu_direct"] = True
-
-    async with app.state.services.http_client.get(search_url, params=params) as resp:
+    async with app.state.services.http_client.get(
+        app.settings.MIRROR_SEARCH_ENDPOINT, params=params,
+    ) as resp:
         if resp.status != status.HTTP_200_OK:
-            if USING_CHIMU:
```
minisbett, line 493: This can be removed because chimu no longer returns a 404 if no map was found

app/api/domains/osu.py
```diff
@@ -1784,13 +1750,10 @@ async def get_osz(
...
 
-    if USING_CHIMU:
-        query_str = f"download/{map_set_id}?n={int(not no_video)}"
-    else:
-        query_str = f"d/{map_set_id}"
+    query_str = f"{map_set_id}?n={int(not no_video)}"
 
     return RedirectResponse(
-        url=f"{app.settings.MIRROR_URL}/{query_str}",
+        url=f"{app.settings.MIRROR_DOWNLOAD_ENDPOINT.strip('/')}/{query_str}",
```
NiceAesth, line 1756: We shouldn't allow for 2 different formats

app/settings.py
```diff
@@ -48,7 +48,8 @@ def read_list(value: str) -> list[str]:
 OSU_API_KEY = os.environ["OSU_API_KEY"]
 
 DOMAIN = os.environ["DOMAIN"]
-MIRROR_URL = os.environ["MIRROR_URL"]
+MIRROR_SEARCH_ENDPOINT = os.environ["MIRROR_SEARCH_ENDPOINT"]
```
NiceAesth, line 51: This is a breaking change. @cmyui how do we handle migrations for stuff like this?

app/api/domains/osu.py
```diff
@@ -551,19 +517,19 @@ def handle_invalid_characters(s: str) -> str:
...
-                Artist=handle_invalid_characters(bmap["Artist"]),
-                Title=handle_invalid_characters(bmap["Title"]),
-                Creator=bmap["Creator"],
-                RankedStatus=bmap["RankedStatus"],
-                LastUpdate=bmap["LastUpdate"],
-                HasVideo=bmap["HasVideo"],
-                diffs=diffs_str,
-                **{DIRECT_SET_ID_SPELLING: bmap[DIRECT_SET_ID_SPELLING]},
-            ),
+        fmstr = DIRECT_SET_INFO_FMTSTR.format(
```
NiceAesth, line 520: Any reason for this change?

NiceAesth (changes requested):

NiceAesth:
Has chimu been tested since?

minisbett:
> Has chimu been tested since?

Kotrik said he fixed the issues with chimu, I will test it tomorrow

app/api/domains/osu.py
minisbett: I don't quite get what you mean

app/api/domains/osu.py
minisbett: Afaik it is programming standard to always do this for compatibility, just like you use path combining tools instead of just concatenating strings

app/settings.py
minisbett: Actually true haven't thought about that

app/api/domains/osu.py
minisbett: You mean that I'm saving it in an extra variable? I think it's just something that happened while debugging, should I revert it?

NiceAesth: ye, was just curious as to why u changed it.  also i assume set id spelling is gone

app/api/domains/osu.py
NiceAesth: the entire project disallows this. it should not be handled here. if you wish to do this, open up a separate PR and strip leading slashes on config READ so you only do it once

app/api/domains/osu.py
minisbett: Yeah chimu had SetId instead of SetID which is a violation of the cheese gull standard and cmyui for some reason rather implemented that than telling kotrik about it but kotrik said he fixed it now

app/api/domains/osu.py
minisbett: I already suggested that but cmyui didn't want to do that tho it would help a lot since there has been like 20 people who always asked for help because they had an issue with the mirrors not working until we eventually found out they just had a / at the end

NiceAesth: either way this should be removed since it has nothing to do with this pr (also if you do end up opening a pr doing this, use rstrip instead)
```suggestion
        url=f"{app.settings.MIRROR_DOWNLOAD_ENDPOINT/{query_str}",
```

app/api/domains/osu.py
NiceAesth: > You mean that I'm saving it in an extra variable? I think it's just something that happened while debugging, should I revert it?

can revert that btw

app/settings.py
minisbett: let him cook

NiceAesth:
changes are ok to me, up to cmyui if he wants to handle `.env` breaking change somehow

NiceAesth (approved):
things to consider:

contains a breaking change in the `.env` config
`/preview/` removed (will now always forward to osu servers)

app/api/domains/osu.py
```diff
@@ -434,20 +434,12 @@ async def lastFM(
...
-USING_CHIMU = "chimu.moe" in app.settings.MIRROR_URL
-USING_NASUYA = "nasuya.xyz" in app.settings.MIRROR_URL
-
-DIRECT_SET_ID_SPELLING = "SetId" if USING_CHIMU else "SetID"
-
 DIRECT_SET_INFO_FMTSTR = (
-    "{{{setid_spelling}}}.osz|{{Artist}}|{{Title}}|{{Creator}}|"
-    "{{RankedStatus}}|10.0|{{LastUpdate}}|{{{setid_spelling}}}|"
-    "0|{{HasVideo}}|0|0|0|{{diffs}}"  # 0s are threadid, has_story,
+    "{SetID}.osz|{Artist}|{Title}|{Creator}|"
```
cmyui, line 438: @minisbett does this work with chimu? (did they fix the inconsistency?)

app/api/domains/osu.py
```diff
@@ -484,48 +469,28 @@ async def osuSearchHandler(
...
-            if USING_CHIMU:
-                # chimu uses 404 for no maps found
-                if resp.status == status.HTTP_404_NOT_FOUND:
-                    return b"0"
-
             return b"-1\nFailed to retrieve data from the beatmap mirror."
 
-        if USING_NASUYA:
-            # nasuya returns in osu!direct format
-            return await resp.read()
```
cmyui, line 502: should still keep this case as it's a significant simplification

minisbett: hmm I don't know, I don't think it has any benefits. The time of quickly converting it is like close to nothing and because of that the extra code is just unnecessary. Just have mirrors use cheesegull and its all good

app/api/domains/osu.py
minisbett: yes I talked with kotrik and he fixed it

app/api/domains/osu.py
cmyui: cheesegull is really not a great protocol lol

minisbett: But if you really plead to, maybe we should instead make a list of "direct direct" mirrors OR what would be better pass a query parameter to the mirror that the mirror has to handle which would be ideal for generalization

minisbett: > cheesegull is really not a great protocol lol

not great but good enough, no reason to swap it out if all mirrors support it rn

cmyui: perhaps best to abstract this into functions, like handle_cheesegull(), handle_osudirect(), etc.

but for another pr

minisbett: we can discuss that tomorrow on the discord then I'll do a PR after this is merged

Calemy:
you guys might wanna change the example url from uk.catboy.best to catboy.best since it's now the main domain to use

minisbett:
> you guys might wanna change the example url from uk.catboy.best to catboy.best since it's now the main domain to use

lol true

app/api/domains/osu.py
Calemy: > should still keep this case as it's a significant simplification if mirrors want to support serialization to osu!direct format

mino supports osu!direct format since it's creation with the "raw" query

i'm still open to rather return a response that uses the osu!api v2 format instead of cheesegull because, seriously why was that even a thing?

https://catboy.best/api/search?raw=1

tsunyoku: when bpy was made, all existing mirrors used exclusively cheesegull. everybody used ripple which only supported that format, so it just continued… i think api v2 format is definitely better, but it’s a breaking change that will likely stop a lot of mirrors being compatible which seems like the opposite goal of this pr

Calemy: if we don't make this breaking change it won't give mirrors a reason to switch away. I would welcome this change and i'm sure i'm not the only one who sees that, even speaking from a mirror perspective. We would be more of a mirror by using the same responses as osu!api v2. It would also profit other developers outside bancho.py

tsunyoku: > if we don't make this breaking change it won't give mirrors a reason to switch away
as much as this is true, we have to consider who is effected by such breaking change. we should support it but not enforce it.

the point of this pr to generalise mirror usage, and introducing a change supported by 2 mirrors (yours, and mine) is really against that. i’d rather use this pr as a “hey mirror developers, pls support api v2 going forward” and THEN we can talk about requiring it’s usage. this is too much destruction.

tsunyoku: yeah i agree we can and should get mirror devs to support it. i just don’t want anything which enforced it’s use until the majority/ all support it

minisbett: tbh I don't see a disadvantage of cheesegull. It's bad design but it is what established. Not only do private servers rely on cheesegull but also many discord bots, websites etc.

minisbett: > i still think this change is necessary in near future to have it more generalized

How would establishing a new standard help with generalization?

tsunyoku: > 

+

tsunyoku: > tbh I don't see a disadvantage of cheesegull. It's bad design but it is what established. Not only do private servers rely on cheesegull but also many discord bots, websites etc.

+1

Calemy: > tbh I don't see a disadvantage of cheesegull. It's bad design but it is what established. Not only do private servers rely on cheesegull but also many discord bots, websites etc.

I politely disagree. The main reason i dislike cheesegull for example is of how unfriendly it is for developers for being supposed to be a mirror for osu. People have to write a completely different parser if they want to add a mirror into their bot as a fallback to osu or vice versa. Also Cheesegull may give less beatmap information than just a simple osu!api v2 return. You agreeing that it's a bad design confuses me when you also say that you don't see any disadvantages.
Giving it "a new standard" also sounds weird if we think about the fact that we following the original standard should be used in the first place. Like i said, if noone changes something, then it never will become a standard and it gets harder to migrate away from it.

Calemy: using the original structure is simply "the standard" we should have been using instead of cheesegull. It isn't new, it's been already there since ever and there is no reason to not use it since we pratically don't *need* Cheesegull.

minisbett: Well reasons not not switch off have been named in this convo. Yes Cheesegull has flaws and there should be a different standard but that's how standards work, it's hard to just change them.

app/commands.py
```diff
@@ -1267,7 +1269,8 @@ async def server(ctx: Context) -> Optional[str]:
             f"{build_str} | uptime: {seconds_readable(uptime)}",
             f"cpu(s): {cpus_info}",
             f"ram: {ram_info}",
-            f"mirror: {mirror_url} | osu!api connection: {using_osuapi}",
+            f"search mirror: {mirror_search_url} | download mirror: {mirror_download_url}",
+            "osu!api connection: {using_osuapi}",
```
cmyui, line 1273:
```suggestion
            f"osu!api connection: {using_osuapi}",
```

app/api/domains/osu.py
cmyui: i don't think it's best practice to program this loosely, but in this context yes it's 99.9_% fine

cmyui (approved):

cmyui:
might end up re-adding the short path for mirrors sending direct osu!direct results

PR #369, opened 2023-01-11
Fix !server command requirements.txt parsing

minisbett:
The requirements.txt has been modified recently and no longer has the simple format package==version. It also now contains lines like 

`-i https://pypi.org/simple` or 
`git+https://github.com/osuAkatsuki/akatsuki-pp-py@abd33445a4f48ddf69c8e7996f7e9ed7340606a0#egg=akatsuki-pp-py`

that need to be handled accordingly.


minisbett:
ffs it messed up branches again

Issue #335, opened 2023-01-07
`!server` command requirements parsing fails


shoei1944:
```
Traceback (most recent call last):
  File "/home/shoei1944/server/app/commands.py", line 2567, in process_commands
    res = await cmd.callback(
  File "/home/shoei1944/server/app/commands.py", line 1367, in server
    requirements_info = "\n".join(
  File "/home/shoei1944/server/app/commands.py", line 1368, in <genexpr>
    " | ".join("{} v{}".format(*pkg.split("==")) for pkg in section)
  File "/home/shoei1944/server/app/commands.py", line 1368, in <genexpr>
    " | ".join("{} v{}".format(*pkg.split("==")) for pkg in section)
IndexError: Replacement index 1 out of range for positional args tuple
[04:44:28PM] <shoei1944 (3)> @ <#osu>: !server
```

![image](https://user-images.githubusercontent.com/119505551/211161418-36cfd003-32c6-45f7-8b43-c26399a956c6.png)

cmyui:
Ahh wow that code is pretty based.. It depends on `requirements.txt` being in pwd as well as it using explicit versioning..

If we really want to keep `!server` showing dependency versions, it may be possible to pull them from the runtime, that would be much better than depending on this file

On a related note, eventually I would really like to remove this line from main.py; which exists because the server stores data in `bancho.py/.data`.
https://github.com/osuAkatsuki/bancho.py/blob/3a9efd82e8e611c50a58d7f3c2f4c0f48484b4df/main.py#L23

PR #370, opened 2023-01-11
fix: `!server` command requirements.txt parsing

minisbett:
The requirements.txt has been modified recently and no longer has the simple format package==version. It also now contains lines like

`-i https://pypi.org/simple` or
`git+https://github.com/osuAkatsuki/akatsuki-pp-py@abd33445a4f48ddf69c8e7996f7e9ed7340606a0#egg=akatsuki-pp-py`

that need to be handled accordingly.
I also wanted to handle that the requirements-dev.txt is used but I don't know how to check for when this is used

EDIT: Switched to using importlib.metadata for getting the runtime dependencies


app/commands.py
```diff
@@ -1256,10 +1256,22 @@ async def server(ctx: Context) -> Optional[str]:
...
     # psutil v5.8.0 | py3rijndael v0.3.3 | uvloop v0.15.2
-    reqs = (Path.cwd() / "requirements.txt").read_text().splitlines()
+    requirements = []
+
+    for line in (Path.cwd() / "requirements.txt").read_text().splitlines():
+        line = line.split(";")[0].strip()
+
+        if line.startswith("git+"):
+            split = line.split("@")[0].split("/")[-2:]
+            requirements.append(f"git:{'/'.join(split)}")
```
NiceAesth, line 1266: Expected behaviour for git packages is to provide the `#egg={pkgname}` info

minisbett: what does the egg mean? I'd like to show both the github user and the repository name, only the repository name won't get you far

NiceAesth: `egg` is the package name of the package when installed

NiceAesth: see: [pip requirements format](https://pip.pypa.io/en/stable/reference/requirements-file-format/)

NiceAesth:
https://github.com/osuAkatsuki/bancho.py/issues/335#issuecomment-1374748884 this should be of interest to you

minisbett:
> [#335 (comment)](https://github.com/osuAkatsuki/bancho.py/issues/335#issuecomment-1374748884) this should be of interest to you

@cmyui I'm not show how you could reliably get all dependencies at runtime without checking the requirements.txt file, any ideas?

app/commands.py
```diff
@@ -1256,10 +1256,15 @@ async def server(ctx: Context) -> Optional[str]:
     # cmyui v1.7.3 | datadog v0.40.1 | geoip2 v4.1.0
     # maniera v1.0.0 | mysql-connector-python v8.0.23 | orjson v3.5.1
     # psutil v5.8.0 | py3rijndael v0.3.3 | uvloop v0.15.2
-    reqs = (Path.cwd() / "requirements.txt").read_text().splitlines()
+    requirements = []
+
+    for dist in importlib.metadata.distributions():
+        requirements.append(f"{dist.name} v{dist.version}")
+    requirements.sort(key=lambda x: x.lower())
```
cmyui, line 1263:
```suggestion
    requirements.sort(key=lambda x: x.casefold())
```

app/commands.py
```diff
@@ -1256,10 +1256,15 @@ async def server(ctx: Context) -> Optional[str]:
     # cmyui v1.7.3 | datadog v0.40.1 | geoip2 v4.1.0
     # maniera v1.0.0 | mysql-connector-python v8.0.23 | orjson v3.5.1
     # psutil v5.8.0 | py3rijndael v0.3.3 | uvloop v0.15.2
-    reqs = (Path.cwd() / "requirements.txt").read_text().splitlines()
+    requirements = []
+
+    for dist in importlib.metadata.distributions():
+        requirements.append(f"{dist.name} v{dist.version}")
```
cmyui, line 1262: i don't think this is worth fixing but the name attribute technically doesn't exist on all distribution classes
```suggestion
        requirements.append(f"{dist.name} v{dist.version}")  # type: ignore
```

cmyui (approved):

PR #371, opened 2023-01-11
Implement modular anticheat

minisbett:
Some idea I had, making a modular anticheat where you can easily define rules using the player and score object.
This can be done by simply defining a new function inside the anticheat module. (see example implementation)

The checks are ran when a score is being submitted, at the time on submission where the pp cap that bpy once had were checked.
If you have any ideas for extending this concept let me know.

The only disatvantage currently is that I can't figure out how to get all functions defined in anticheat.py realiably. Currently I am using `inspect.getmembers(sys.modules[__name__], inspect.isfunction)` with the flaw that it also returns imported functions. Therefore, I currently avoid importing any functions (e.g. `from app.logging import log`)

This anticheat essentially allows every server to easily customize their rules on what should auto-restrict without having to go into the submission and either define their rules in that 400 line beast or creating this class themselves manually.


app/usecases/anticheat.py
```diff
@@ -0,0 +1,51 @@
...
+# The response is a string that is either None or the restriction reason.
+#
+# Here is an example for a pp cap rule:
+async def std_ppcap(player: Player, score: Score) -> Optional[str]:
+
+    if score.mode != GameMode.VANILLA_OSU:
+        return
+
+    if score.pp >= 1500:
+        return "Exceeded std!vn pp cap of 1500p"
```
7ez, line 51: pp cap of 15 qjid

app/usecases/anticheat.py
```diff
@@ -0,0 +1,51 @@
...
+from app.objects.score import Score
+
+# DO NOT import functions directly, it'll mess up inspect.getmembers
+
+# Add checks here that you'd like to have disabled, but not removed from the code entirely
+disabled_checks = []
+
+
+async def run_anticheat_checks(player: Player, score: Score):
+    checks = dict(inspect.getmembers(sys.modules[__name__], inspect.isfunction))
```
cmyui, line 22: core idea is good (i've had similar thought before) but implementation is a bit weird here

better to make this either an (optional) separate service entirely, or at least a module that's under gitignore

PR #372, opened 2023-01-11
feat: v2 clans api (get calls only)

cmyui:


alowave223 (changes requested):
We need to fetch clan members too

cmyui:
> We need to fetch clan members too

true, but this will be a separate api, the endpoints can be /clans/{clan_id}/members but it can be implemented in a separate file to fit nicely with the codebase

PR #373, opened 2023-01-11
feat: v2 maps api (get calls only)

cmyui:


alowave223 (changes requested):
We should add most played maps on a server and by a single player as it is used by guweb atm in order to move guweb from v1 to v2

cmyui:
> We should add most played maps on a server and by a single player as it is used by guweb atm in order to move guweb from v1 to v2

this will be a separate api

app/api/v2/models/maps.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+# input models
+
+
+# output models
+
+
+class Map(BaseModel):
+    server: Literal["osu!", "private"]
```
cmyui, line 16: this should be moved down to match db table

app/api/v2/maps.py
```diff
@@ -0,0 +1,75 @@
...
+from app.api.v2.common.responses import Success
+from app.api.v2.models.maps import Map
+from app.repositories import maps as maps_repo
+
+router = APIRouter()
+
+
+@router.get("/maps")
+async def get_maps(
+    server: str,
```
cmyui, line 20: this should be moved down to match db table; also should test if typing.Literal can/should be used here

cmyui: avoiding for now - there are other similar things happening elsewhere, might standardize these to use enums later.

PR #380, opened 2023-01-13
docs: add issue and pr templates

NiceAesth:


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,11 @@
+blank_issues_enabled: false
```
NiceAesth, line 1: Are you ok with this? Personally I would prefer if we started moving towards discussions for feature requests

.github/pull_request_template.md
```diff
@@ -0,0 +1,12 @@
...
+    - Titles should follow [semantic commit message format](https://sparkbox.com/foundry/semantic_commit_messages)
+-->
+
+# Describe your changes
+
+## Related Issues / Projects
+
+## Checklist
+- [ ] The changes pass pre-commit checks
+- [ ] The changes follow coding style
```
NiceAesth, line 12: todo: add checklist for passing unit testing and following community guidelines

.github/ISSUE_TEMPLATE/config.yml
cmyui: does this mean people can't open an issue with no description? (could they use a custom description that doesn't fit our template still?)

NiceAesth: yeah, it means people would have to follow the template when opening issues. can remove this if needed

NiceAesth: ideally we could just add a new template for other issues which you can type anything in

NiceAesth: added misc issue template, hopefully that should be enough

NiceAesth:
https://github.com/NiceAesth/test-templates

cmyui (approved):

cmyui:
thanks! this is much nicer imo

Issue #379, opened 2023-01-13
Bring repository up to community standards


NiceAesth:
- [x] Add code of conduct (https://github.com/osuAkatsuki/bancho.py/pull/383)
- [x] Add security policy (https://github.com/osuAkatsuki/bancho.py/pull/381)
- [x] Add issue templates (https://github.com/osuAkatsuki/bancho.py/pull/380)
- [x] Add pull request templates (`feat`, `bug`, etc.) (https://github.com/osuAkatsuki/bancho.py/pull/380)
- [ ] Add contributing guidelines 
- [ ] Add code style guidelines (documentation in general)
- [x] Document breaking changes in wiki
- [x] Create language picker in wiki

NiceAesth:
@jacksonisiah

cmyui:
nice work on this, so quickly

PR #381, opened 2023-01-13
docs: add `SECURITY.md`

NiceAesth:
# Describe your changes

Added security policy

## Related Issues / Projects

#379 

## Checklist
- [x] The changes pass pre-commit checks

### DIFF
--- a/.github/SECURITY.md
+++ b/.github/SECURITY.md
@@ -1,0 +1,12 @@
+# Security Policy
+
+## Supported Versions
+
+bancho.py is in a volatile state, therefore we only provide support for the latest version. It is the responsibility of the user to remain up-to-date.
+
+## Reporting a Vulnerability
+
+If you notice any vulnerabilities, pleae contact us urgently on
+[Discord](https://discord.gg/ShEQgUx) or via security advisories in the repository.
+
+Depending on the issue, a resolution will be attempted in up to a week.
--- a/.github/SECURITY.md
+++ b/.github/SECURITY.md
@@ -2,11 +2,13 @@
 
 ## Supported Versions
 
-bancho.py is in a volatile state, therefore we only provide support for the latest version. It is the responsibility of the user to remain up-to-date.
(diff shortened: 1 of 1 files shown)


.github/SECURITY.md
```diff
@@ -0,0 +1,12 @@
+# Security Policy
+
+## Supported Versions
+
+bancho.py is in a volatile state, therefore we only provide support for the latest version. It is the responsibility of the user to remain up-to-date.
```
cmyui, line 5: hmm i'm not sure about this, i generally try to provide support to all versions "within a reasonable timeframe", such as ~6 months - 1 year. if they have made modifications to their codebase, i ask that they provide their source code in a git (or alike) repository so that i can assess whether or not they've made any changes that affect what they're finding issues with.

.github/SECURITY.md
```diff
@@ -0,0 +1,12 @@
...
+
+## Reporting a Vulnerability
+
+If you notice any vulnerabilities, pleae contact us urgently on
+[Discord](https://discord.gg/ShEQgUx) or via security advisories in the repository.
```
cmyui, line 10: would prefer email directly for reporting security issues - cmyuiosu@gmail.com

.github/SECURITY.md
```diff
@@ -0,0 +1,12 @@
...
+
+Depending on the issue, a resolution will be attempted in up to a week.
```
cmyui, line 12: i think it's better to just say it will be prioritized according to severity, rather than giving a timeframe explicitly

.github/SECURITY.md
NiceAesth: sure

.github/SECURITY.md
NiceAesth: alright

.github/SECURITY.md
NiceAesth: sure, i do think that we should not guarantee support for forks though as that is a complete nightmare. check to see if what i added works for you

cmyui (approved):

PR #385, opened 2023-01-14
feat: add anticheat service

minisbett:
# Describe your changes

This PR adds bancho.py's optional modular anticheat (or BOMA in short). It is written in C# with dotnet 7 and the PR already includes readme instructions on how to set everything up and added a script for installing dotnet on Ubuntu 18, 20 and 22 since it requires adding Microsoft's package feed.

Bancho.py publishes a json object containing all useful score data (including the beatmap and player) to an AMQP service on score submission, to be specific right after the score has been added to the database and before any modifications to the player stats are done in order to publish informations as if they were before the score got submitted.
We use RabbitMQ for that. As per readme instructions, you download docker.io and create a docker container for rabbitmq which can then be started and stopped. 

As cmyui suggested, the implemention of publishing the score upon submission can be made a bit broader to also include non-anticheat consumers on the query.  The connection details (host and port) for RabbitMQ are configured in the .env file. There is also an option to enable the RabbitMQ publishing because not every server might use it. By default it is disabled.

In bancho.py, the pip package `aiopika` is used to communicate with RabbitMQ. The service is initialized and shut down the same way bancho.py handles redis and mysql. I added `app/usecases/rabbitmq.py` which handles publishing the data.
There also is a function for serializing the score object to json which might look bloated at first but after trying various ways to serialize the data I found that this is the best way to have control over what's being published.

The anticheat is kept pretty simple. For logging, I chose to follow bancho.py's output and color scheme. The application loads the config.json on startup which contains the RabbitMQ information, as well as more anticheat-related settings.
Then it initializes a `ScoreQueue` object, a wrapper around .NET's queue that subscribes to the queue and populates the underlying queue object with incoming score objects that are being deserialized using Newtonsoft.Json and the models found in `src/Models`.

After that is done, all checks stored in `src/Checks` are loaded using reflection.
All checks implement the abstract `Check` class as an interface for the checks. The anticheat processor passes the consumed score as a parameter to the `PerformCheck` method. From there, checks can be done to to validate the score.
The abstract class provides helper methods/properties for returning the `CheckResult` object. For example, you can request a restriction via `return Restrict("reason")` or tell the anticheat to take no action using `return NoAction`.

```cs
public override CheckResult PerformCheck(Score score)
{
    if (score.Mode == GameMode.VN_STD && score.PP >= 1500)
        return Restrict("Exceeded the PP limit of 1500pp");

    return NoAction;
    }
```
This checks whether the user did a PP play of more than or equal to 1500pp on the vanilla standard gamemode and if so, restricts the user with the provided reason.

After these checks are instantiated, the `AnticheatProcessor` is being initialized that contains the anticheat's lifetime cycle. It dequeues the score object, does a non-performant heavy check on whether the score is eligible for running through the checks that might be performance-intensive (e.g. user is whitelisted or restricted) after which the real checks are performed.

The checks themselves return a `CheckResult` object which contains the type of action the anticheat should react with (currently only `None` or `Restrict` are implemented), as well as additional properties for them, like the restriction message. If a restriction is requested, further checks are not being done to save resources.

The anticheat can be started by going into the anticheat folder and running `./run.sh`. The script makes sure `dotnet run` is executed with the correct arguments. `./run.sh` runs the anticheat in the debug configuration (which is different from the debug in the `config.json` file which only enables debug logging).

I plan on adding more functionality to the anticheat soon (e.g. parsing the replay data into frames that can be processed or adding caches for players so the anticheat can remember certain flags or other kinds of information about the player). But for no, I want to keep the anticheat's state on the initial PR as simple as possible.

## Checklist
- [X] The changes pass pre-commit checks (`make lint`)
- [X] The changes follow coding style


NiceAesth:
pre-commit.ci autofix

app/api/init_api.py
```diff
@@ -131,6 +132,14 @@ async def on_startup() -> None:
         await app.state.services.database.connect()
         await app.state.services.redis.initialize()
 
+        if app.settings.RABBITMQ_ENABLED:
+            app.state.services.amqp = await aio_pika.connect_robust(
+                host=app.settings.RABBITMQ_HOST,
+                port=app.settings.RABBITMQ_PORT,
+            )
+
+        app.state.services.amqp_channel = await app.state.services.amqp.channel()
```
cmyui, line 141: should this be indented? looks like an err with `app.state.services.amqp` undefined otherwise

app/api/init_api.py
```diff
@@ -162,6 +171,10 @@ async def on_shutdown() -> None:
         await app.state.services.database.disconnect()
         await app.state.services.redis.close()
 
+        if app.state.services.amqp is not None:
+            await app.state.services.amqp_channel.close()
+            await app.state.services.amqp.close()
```
cmyui, line 176: should this be `if app.settings.RABBITMQ_ENABLED`?

app/usecases/rabbitmq.py
cmyui, line 1: i think `app/usecases/anticheat.py` is a bit better, thoughts?

ideally we'd have it in app/usecases/scores probably within the submission function, and this would be an adapter

app/api/init_api.py
minisbett: Both would work but generally I would say this condition makes more sense, checking if the object you close is not null instead of checking if the object is technically initialized in the first place.

app/api/init_api.py
minisbett: yup, you're right

app/usecases/rabbitmq.py
minisbett: The idea was that this rabbitmq thing is not just for the anticheat, but instead can be used by any 3rd-party tool to receive all submitted scores.

cmyui: yeah kk it's more like an adapter

cmyui: np to leave it as-is

PR #388, opened 2023-01-15
feature: Implement authentication for API v2 based on OAuth2 potocol.

alowave223:
# Describe your changes
Created authorization for third-party clients based on the OAuth2 protocol standard.
## Related Issues / Projects
[project](feature: Implement authentication for API v2 based on OAuth2 potocol.)
## Checklist

### DIFF
--- a/app/api/v2/__init__.py
+++ b/app/api/v2/__init__.py
@@ -1,11 +1,46 @@
 from __future__ import annotations
 
+from typing import Any
+
 from fastapi import APIRouter
+from fastapi import Depends
+from fastapi import HTTPException
+from fastapi import status
+
+from app.api.v2.common.oauth import OAuth2Scheme
+from app.repositories import access_tokens as access_tokens_repo
+
+
+oauth2_scheme = OAuth2Scheme(
+    authorizationUrl="/v2/oauth/authorize",
+    tokenUrl="/v2/oauth/token",
+    refreshUrl="/v2/oauth/refresh",
+    scheme_name="OAuth2 for third-party clients.",
+    scopes={
+        "public": "Access endpoints with public data.",
+        "identify": "Access endpoints with user's data.",
+        "admin": "Access admin endpoints.",
+    },
+)
+
+
+async def get_current_client(token: str = Depends(oauth2_scheme)) -> dict[str, Any]:
+    """Look up the token in the Redis-based token store"""
+    access_token = await access_tokens_repo.fetch_one(token)
+    if not access_token:
+        raise HTTPException(
+            status_code=status.HTTP_401_UNAUTHORIZED,
+            detail="Not authenticated",
+            headers={"WWW-Authenticate": "Bearer"},
+        )
+    return access_token
+
 
 from . import clans
 from . import maps
 from . import players
 from . import scores
+from . import oauth
 
 apiv2_router = APIRouter(tags=["API v2"], prefix="/v2")
 
@@ -13,3 +48,4 @@ apiv2_router.include_router(clans.router)
 apiv2_router.include_router(maps.router)
 apiv2_router.include_router(players.router)
 apiv2_router.include_router(scores.router)
+apiv2_router.include_router(oauth.router)
--- a/app/api/v2/common/oauth.py
+++ b/app/api/v2/common/oauth.py
@@ -1,0 +1,85 @@
+from __future__ import annotations
+
+import base64
+from typing import Optional
+from typing import Union
+
+from fastapi import Request
+from fastapi import status
+from fastapi.exceptions import HTTPException
+from fastapi.openapi.models import OAuthFlows as OAuthFlowsModel
+from fastapi.security import OAuth2
+from fastapi.security.utils import get_authorization_scheme_param
+
+
+class OAuth2Scheme(OAuth2):
+    def __init__(
+        self,
+        authorizationUrl: str,
+        tokenUrl: str,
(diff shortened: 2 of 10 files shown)


app/api/v2/models/oauth.py
```diff
@@ -0,0 +1,22 @@
+from __future__ import annotations
```
alowave223, line 1: This import is breaking FastApi's `depends`. Related issue: https://github.com/tiangolo/fastapi/issues/1654

app/api/v2/common/oauth.py
```diff
@@ -0,0 +1,85 @@
+from __future__ import annotations
```
alowave223, line 1: This import is breaking FastApi's `depends`. Related issue: https://github.com/tiangolo/fastapi/issues/1654

app/api/v2/__init__.py
```diff
@@ -1,15 +1,51 @@
...
+    scheme_name="OAuth2 for third-party clients.",
+    scopes={
+        "public": "Access endpoints with public data.",
+        "identify": "Access endpoints with user's data.",
+        "admin": "Access admin endpoints.",
+    },
+)
+
+
+async def get_current_client(token: str = Depends(oauth2_scheme)) -> dict[str, Any]:
```
alowave223, line 27: I tried to do this with the inner function but can't pass the arguments to the function in `Depends`.

cmyui:
hey, there are quite a few typing errors still, i suspect your vscode is using `type_checking_mode: off` in your preferences

you can open preferences with `ctrl+,` and set `type checking mode` to `basic` to have vs code redline any of these errors

image

e.g.
image

image

app/api/v2/oauth.py
```diff
@@ -66,11 +66,11 @@ async def token(
     grant_type: str = Form(),
     client_id: int = Form(default=None),
     client_secret: str = Form(default=None),
-    auth_credentials: Optional[dict[str, Union[str, int]]] = Depends(
+    auth_credentials: Optional[dict[str, Any]] = Depends(
```
cmyui: @alowave223 typing the possible values of a dict is usually a bad idea imo - if you're going to try, consider using a [typing.TypedDict](https://docs.python.org/3/library/typing.html#typing.TypedDict)

cmyui:
there are also quite a few cases like this `state: str = Query(default=None)`, where the variable is typed as `T` but at runtime can actually be of type `Optional[T]` due to the default case

app/api/v2/common/oauth.py
cmyui: yeah i ran into this a long while ago at work, nice work actually tracking it down lol, that's not an easy error

app/api/v2/common/oauth.py
```diff
@@ -0,0 +1,85 @@
...
+                )
+            else:
+                return None
+        return param
+
+
+# https://developer.zendesk.com/api-reference/sales-crm/authentication/requests/#client-authentication
+def get_credentials_from_basic_auth(
+    request: Request,
+) -> Optional[dict[str, Union[str, int]]]:
```
cmyui, line 66: better to use a TypedDict here, something like this
```suggestion

from typing import TypedDict

class BasicAuthCredentials(TypedDict):
    client_id: int
    client_secret: str

# https://developer.zendesk.com/api-reference/sales-crm/authentication/requests/#client-authentication
def get_credentials_from_basic_auth(
    request: Request,
) -> Optional[BasicAuthCredentials]:
```

app/api/v2/oauth.py
```diff
@@ -0,0 +1,207 @@
+""" bancho.py's v2 apis for interacting with clans """
```
cmyui, line 1: some docstrings still out of date

app/api/v2/__init__.py
cmyui: np

cmyui (changes requested):
left a review with some changes requested

overall i'd like to also reduce usage of typing.Union - better to constrain the variance of types when they come into our ecosystem so that we do not have to deal with this complexity ourselves

nice work so far! & sorry for the very late pr review lol

app/settings.py
```diff
@@ -84,4 +84,4 @@ def read_list(value: str) -> list[str]:
 ## WARNING touch this if you know how
 ##          the migrations system works.
 ##          you'll regret it.
-VERSION = "4.8.1"
+VERSION = "4.8.1
```
cmyui, line 87:
```suggestion
VERSION = "4.8.1"
```

cmyui:
started work, a couple of things stand out:
1. since this is meant to be an api for frontend clients, i think we should use jwt and avoid storing raw access tokens on our server. we can avoid the need to store raw access tokens, and instead store only identifiers to allow for token revocation by server staff.
2. i'm not certain whether it's best to persist access tokens vs. refresh tokens vs. both -- i'll have to look a bit deeper into the oauth 2.0 standard to see if there's a reason to prefer one over the other.
3. there are some typing/code implementation inconsistencies -- shouldn't be very major, the two above issues are more fundamental questions/blocking in nature.

cmyui:
> 2. i'm not certain whether it's best to persist access tokens vs. refresh tokens vs. both -- i'll have to look a bit deeper into the oauth 2.0 standard to see if there's a reason to prefer one over the other.

It may also be worth storing other things, such as authorization grants and failed authorization attempts.

PR #396, opened 2023-01-31
fix: pre-commit conflicting deps

NiceAesth:
# Describe your changes

`blacken-docs` was requesting a version of black that is out of date.

## Related Issues / Projects

This PR fixes checks, should be merged immediately.

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,7 +33,6 @@ repos:
     rev: 1.13.0
     hooks:
     -   id: blacken-docs
-        additional_dependencies: [black==21.11b1]
 
 default_language_version:
     python: python3.9


.pre-commit-config.yaml
```diff
@@ -33,7 +33,6 @@ repos:
     rev: 1.13.0
     hooks:
     -   id: blacken-docs
-        additional_dependencies: [black==21.11b1]
```
cmyui, line 36: can't we just set it to 22.12.0 (like is used above)? this *is* an additional dependency that's required for this hook, i don't think we should remove it outright

NiceAesth: i see VERY little point in adding the dependency there when black itself is already in the same file as a dependency. this is just a liability that we have to maintain on every update at this point

cmyui: fair

cmyui (approved):

.pre-commit-config.yaml
NiceAesth: i would be fine with adding it back if pre-commit ci would handle it but it doesn't sadly

PR #398, opened 2023-01-31
fix: account for sv2 in mania acc calculation

NiceAesth:
# Describe your changes

[Mania uses a different formula to calculate accuracy on SV2](https://osu.ppy.sh/wiki/en/Gameplay/Accuracy#osu!mania). Currently, SV2 scores have an incorrect accuracy inserted into the database. This change should not affect pp calculation since that uses a different acc formula.

## Related Issues / Projects

Discussed in: https://discord.com/channels/748687781605408908/790425632017219605/1070015841971740712

## Checklist


tsunyoku (approved):

tsunyoku:
this is an issue for score v2 on std too lmao, that's gonna be a way bigger issue that i never thought about

NiceAesth:
on std the difference would be that sliders will be added to the objects that can have accuracy, correct? (also for people who have lazer in mind, this would also need to be enabled for lazer scores unless classic mod is enabled and has slideracc off)

tsunyoku:
> on std the difference would be that sliders will be added to the objects that can have accuracy, correct?
yeah, might not actually be an issue.. i keep forgetting the 300 etc. count comes from the client, i was imagining that we would have to replay simulate to find what sliders 100s were set on but i'm just getting ahead of myself

NiceAesth:
it is an issue, but only pp-calc side. we use weighted accuracy to calculate pp which accounts for objects which cannot have accuracy separately

cmyui (approved):

PR #414, opened 2023-03-10
fix: ignore existing environments; do not expose db by default

NiceAesth:
## Checklist


NiceAesth:
@7mochi are you going to push your ver?

7mochi:
> @7mochi are you going to push your ver?

yeah, I just forgot and now I don't have Internet until 6pm (11am here)

7mochi:
Done

Dockerfile
```diff
@@ -18,6 +18,8 @@ COPY Makefile Pipfile Pipfile.lock ./
 RUN python3.9 -m pip install -U pip==23.0.1 setuptools==67.6.0 pipenv==2023.2.18
 RUN make install
 
+# create data directory
+RUN mkdir /home/bpyuser/.data
```
cmyui, line 22: nit
```suggestion
RUN mkdir /home/bpyuser/.data

```

docker-compose.yml
```diff
@@ -38,6 +44,7 @@ services:
     tty: true
     volumes:
       - data:/home/bpyuser/.data
+      - .env:/home/bpyuser/.env:ro
```
cmyui, line 47: seems a bit weird imo, i think better to map the env variables into the container using `env` so they're available from anything running inside it

NiceAesth: pretty sure bpy will not be happy with not having a .env file

cmyui:
hey @7mochi, has anything changed that might make this need updates?

7mochi:
> hey @7mochi, has anything changed that might make this need updates?

I don't think so, there are other changes I would like to implement to try to make the configuration easier (With and without docker) but I'll do it in another pr

docker-compose.yml
cmyui: it'll be fine - it takes .env file values first, then environment variables 2nd

ideally everything should be in the environment of the running process "globally" rather than only in the file

cmyui: can merge for now anyways

cmyui (approved):

PR #416, opened 2023-03-15
fix: osu erroring on map change when mode is >= 4

7ez:
# Describe your changes

i've also removed the unnecessary GameMode() conversions as mode on match is typehinted as int anyway

## Checklist


tsunyoku (approved):

cmyui:
the removal of the casts wasn't correct here :/

in the parsed match_data it's int, but in the internal objects they're GameMode, so it needs conversion

image

image

cmyui:
will fix in upcoming pr

PR #419, opened 2023-03-30
chore: Add bidirectional sync between .github/docs/wiki and wiki

7mochi:
Depends on: https://github.com/osuAkatsuki/bancho.py/pull/460

# Describe your changes
This workflow syncs the `.github/docs/wiki` folder with the repo wiki and vice versa

## Checklist


cmyui:
is it possible to do bi-directional sync?

7mochi:
> is it possible to do bi-directional sync?

It is, i modified the workflow to make use of this [action](https://github.com/newrelic/wiki-sync-action). That action is enough to make the sync bidirectional but currently I think it has a bug when it tries to push from the repo to the wiki:
image
so for that job I have left the first method of sync that i pushed and for the other one I'm using the action

cmyui:
@NiceAesth thoughts on this?

NiceAesth:
i find this solution a bit moot as editing the wiki without this will only pose a problem to those very unfamiliar with git / github. i would really much rather use the action and the error you are getting is almost 100% because of misconfiguration on your side (probably wiki didnt exist on the repo or something, check the repo for the action for more info)

7mochi:
> i find this solution a bit moot as editing the wiki without this will only pose a problem to those very unfamiliar with git / github. i would really much rather use the action and the error you are getting is almost 100% because of misconfiguration on your side (probably wiki didnt exist on the repo or something, check the repo for the action for more info)

you're right, I solved the issue in my test repo and now I'm using the action for both jobs. for the first job it's necessary to generate a personal access token with the name `BPY_REPO_TOKEN`, if I use the `GITHUB_TOKEN` it gives me the following error
image
idk if this is also a misconfiguration on my part, if so we could also use the `GITHUB_TOKEN` for the first job

[this is my test repo](https://github.com/7mochi/wiki-test) where you can see how the pr works

NiceAesth:
does this work fine now?

7mochi:
> does this work fine now?

yes

tsunyoku (approved):

NiceAesth:
I have one question actually, before this gets merged. How does this handle existing wikis? Will it create the aforementioned wiki directory or will it wipe the existing content?

7mochi:
> I have one question actually, before this gets merged. How does this handle existing wikis? Will it create the aforementioned wiki directory or will it wipe the existing content?

I haven't tested that case tbh. i'll do it in a few hours in another repo

7mochi:
> > > I have one question actually, before this gets merged. How does this handle existing wikis? Will it create the aforementioned wiki directory or will it wipe the existing content?
> > 
> > 
> > I haven't tested that case tbh. i'll do it in a few hours in another repo
> 
> So, has in a few hours happened? sweat_smile

Sorry I've been a little busy :tf: after my meeting at work today I'll test it

NiceAesth:
oh ffs

PR #422, opened 2023-04-06
interactive ui for online players & matches on GET /c.ppy.sh page

arily:
# Describe your changes
display more information in the "bancho" page, key to judge if it's safe to restart server without login using osu client.

image
image
image

## Checklist
- [ ] The changes pass pre-commit checks (`make lint`)

### DIFF
--- a/app/api/domains/cho.py
+++ b/app/api/domains/cho.py
@@ -170,8 +170,8 @@ matches:
       f'''{(ON_GOING if m.in_progress else IDLE).ljust(max_status_length)} ({str(m.id).rjust(match_id_max_length)}): {m.name}
 -- '''
       + f"{new_line}-- ".join([
-            f'{BEATMAP.rjust(max_properties_length)}: {m.map_name}',
-            f'{HOST.rjust(max_properties_length)}: <{m.host.id}> {m.host.safe_name}'
+            f'{BEATMAP.ljust(max_properties_length)}: {m.map_name}',
+            f'{HOST.ljust(max_properties_length)}: <{m.host.id}> {m.host.safe_name}'
         ])
     ,
     matches,
--- a/app/api/domains/cho.py
+++ b/app/api/domains/cho.py
@@ -88,7 +88,7 @@ async def bancho_http_handler():
     matches = list(
         filter(lambda match: isinstance(match, Match), app.state.sessions.matches),
     )
-    players = list(filter(lambda p: not p.bot_client, app.state.sessions.players))
+    players = [p for p in app.state.sessions.players if not p.bot_client]
 
     packets = app.state.packets["all"]
     return HTMLResponse(
--- a/app/api/domains/cho.py
+++ b/app/api/domains/cho.py
@@ -121,19 +121,9 @@ async def bancho_list_user():
 <!DOCTYPE html>
 <body style="font-family: monospace;  white-space: pre-wrap;"><a href="/">back</a>
 users:
-{new_line.join(
-    map(
-        lambda p: f"({p.id:>{id_max_length}}): {p.safe_name}",
-        players,
-    ),
-)}
+{new_line.join([f"({p.id:>{id_max_length}}): {p.safe_name}" for p in players])}
 bots:
-{new_line.join(
-    map(
-        lambda p: f"({p.id:>{id_max_length}}): {p.safe_name}",
-        bots,
-    ),
-)}
+{new_line.join(f"({p.id:>{id_max_length}}): {p.safe_name}" for p in bots)}
 </body>
 </html>""",
     )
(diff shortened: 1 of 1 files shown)


cmyui:
i knew this day would come

frontend improvements to bancho.py LOL

app/api/domains/cho.py
```diff
@@ -84,20 +84,100 @@
 @router.get("/")
 async def bancho_http_handler():
     """Handle a request from a web browser."""
+    new_line = "\n"
+    matches = list(
+        filter(lambda match: isinstance(match, Match), app.state.sessions.matches),
+    )
```
cmyui, line 90: are they ever not of instance `Match`?

app/api/domains/cho.py
```diff
@@ -84,20 +84,100 @@
...
+    players = list(filter(lambda p: not p.bot_client, app.state.sessions.players))
```
cmyui, line 91:
```suggestion
    players = [p for p in app.state.sessions.players if not p.bot_client]
```

cmyui: generally more concise

app/api/domains/cho.py
arily: [there's None](https://github.com/arily/bancho.py/blob/9078c4f68e22d6b8f096baeef9d918d48803e2e4/app/objects/collections.py#L175)

app/api/domains/cho.py
```diff
@@ -84,20 +84,100 @@
...
+    )
+
+
+@router.get("/online")
+async def bancho_list_user():
+    """see who's online"""
+    new_line = "\n"
+    players = list(filter(lambda p: not p.bot_client, app.state.sessions.players))
+    bots = list(filter(lambda p: p.bot_client, app.state.sessions.players))
+    id_max_length = len(str(max(p.id for p in app.state.sessions.players)))
```
cmyui, line 117: same can be done for these

app/api/domains/cho.py
cmyui: oh, probably better to check for that

cmyui:
```py
matches = [m for m in app.state.sessions.matches if m is not None]
```

app/api/domains/cho.py
cmyui: and the others later on

app/api/domains/cho.py
arily: f-strings below will throw error if None is not filtered out, tested locally

app/api/domains/cho.py
```diff
@@ -84,20 +84,100 @@
...
-                "<br>".join([f"{packet.name} ({packet.value})" for packet in packets]),
-            ),
-        ).encode(),
+        f"""
+<!DOCTYPE html>
+<body style="font-family: monospace;  white-space: pre-wrap;"><a href="/">back</a>
+users:
+{new_line.join(
+    map(
+        lambda p: f"({str(p.id).rjust(id_max_length)}): {p.safe_name}",
```
cmyui, line 126: if i remember correctly, i think this should work as well for the `f"{str(x).rjust(y)}"` cases, but i don't particularly care either way
```suggestion
        lambda p: f"({p.id:>{id_max_length}}): {p.safe_name}",
```

app/api/domains/cho.py
cmyui: yea you 100% need to remove the nones, the `if m is not None` does the filtering in my example

app/api/domains/cho.py
arily: LGTM

arily: how to do a ljust? <: ?

cmyui: (there are many more cases of this though - my only request is to pick one style and use it consistently)

cmyui: almost - `:<`

app/api/domains/cho.py
arily: fixed

app/api/domains/cho.py
cmyui: in python `:` is the delimiter between the data and the format spec

e.g.
```py
balance = 123.4567
print(f"Balance: ${123:.2f}") # here, the format spec is `.2f`, for "2 decimal float"
> 123.45
```

cmyui: more format specs here https://www.w3schools.com/python/ref_string_format.asp

app/api/domains/cho.py
arily: updated

app/api/domains/cho.py
arily: updated

arily:
I think you can squash the commits. It's a little bit cluttered rn

arily:
I think it's ready now

cmyui (approved):

cmyui:
thanks for the pr!

PR #423, opened 2023-04-08
change kitsu to osu.direct

HorizonCode:
since Kitsu got rebranded to osu.direct, the url changed

### DIFF
--- a/app/objects/beatmap.py
+++ b/app/objects/beatmap.py
@@ -37,7 +37,7 @@ async def api_get_beatmaps(**params: Any) -> Optional[list[dict[str, Any]]]:
     """\
     Fetch data from the osu!api with a beatmap's md5.
 
-    Optionally use Kitsu's API if the user has not provided an osu! api key.
+    Optionally use osu.direct's API if the user has not provided an osu! api key.
     """
     if app.settings.DEBUG:
         log(f"Doing api (getbeatmaps) request {params}", Ansi.LMAGENTA)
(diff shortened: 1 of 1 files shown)


7ez:
real osu direct

app/objects/beatmap.py
```diff
@@ -47,8 +47,8 @@ async def api_get_beatmaps(**params: Any) -> Optional[list[dict[str, Any]]]:
         url = "https://old.ppy.sh/api/get_beatmaps"
         params["k"] = str(app.settings.OSU_API_KEY)
     else:
-        # https://doc.kitsu.moe/
-        url = "https://kitsu.moe/api/get_beatmaps"
+        # https://osu.direct/doc
+        url = "https://osu.direct/api/get_beatmaps"
```
cmyui, line 51: is it backwards compatible? seems like https://kitsu.moe/api/get_beatmaps goes to https://osu.direct/home

cmyui (approved):

app/objects/beatmap.py
cmyui: this should be pretty easy to make 100% backwards compatible with some rerouting rules,
`https://kitsu.moe/$1` -> `https://osu.direct/$1`

tsunyoku: they seemed to have hard redirected to /home always, it broke akatsuki. this sucks :(

cmyui: @HorizonCode we recently had this problem when moving from akatsuki.pw to akatsuki.gg

if you're using cloudflare, consider setting up page rules like this for this (and future domain switches)
image

HorizonCode: i will add that 👍

PR #426, opened 2023-04-11
Fix issue with v1/get_replay API route returning 0 byte files

def750:
# Describe your changes
Fixed an issue related to `v1/get_replay` API route, which have lead to `TypeError: 'bytearray' object is not an iterator` error.

## Checklist

### DIFF
--- a/app/api/v1/api.py
+++ b/app/api/v1/api.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 import hashlib
 import struct
+from io import BytesIO
 from pathlib import Path as SystemPath
 from typing import Literal
 from typing import Optional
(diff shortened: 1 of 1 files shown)


0i8:
would you receive this bug at random?

0i8:
> > would you receive this bug at random? i've never encountered this
> 
> Not at random, all the time. Even on clean unedited gulag install

this is weird, i've never experienced this ever after i added headers to every replay

cmyui:
Hey please try fastapi.responses.FileResponse

cmyui (changes requested):
left review

cmyui:
actually it probably only allows path i guess

cmyui:
nvm

PR #427, opened 2023-04-12
fix typing errors in performance calc from #348

cmyui:
# Describe your changes
TypedDicts were not kept up to date with code in:
- #348

## Checklist


app/api/v1/api.py
```diff
@@ -181,9 +181,10 @@ async def api_calculate_pp(
     )
 
     # "Inject" the accuracy into the list of results
-    if len(results) > 1:
```
cmyui, line 184: @minisbett what was the reason for the length check?

PR #428, opened 2023-04-16
fix: check for empty beatmap file response

arily:
# Describe your changes
current implementation has a flaw to let through empty .osu files:
image
https://github.com/osuAkatsuki/bancho.py/blob/8142763a95a76c734f7a031a0767663ec381a191/app/objects/beatmap.py#L78
this check will fail and let through unknown file to local cache.

I left some comments in the new impl as I saw at least one caller had the type `Any | str` of `bmap_md5` , and remove the explicity check if it's irrelevant.

## Checklist


app/objects/beatmap.py
```diff
@@ -57,33 +57,49 @@ async def api_get_beatmaps(**params: Any) -> Optional[list[dict[str, Any]]]:
     return None
 
 
+KNOWN_BAD_FILES_MD5 = ["d41d8cd98f00b204e9800998ecf8427e"]  # empty file
```
NiceAesth, line 60: absolutely no way i am allowing this through to the codebase. no magic values please.

NiceAesth (changes requested):
see comments

app/objects/beatmap.py
```diff
@@ -57,33 +57,49 @@ async def api_get_beatmaps(**params: Any) -> Optional[list[dict[str, Any]]]:
...
+    if file_md5 == bmap_md5:
+        return True
+    # need to get the file from the osu!api
+    if app.settings.DEBUG:
+        log(f"Doing osu!api (.osu file) request {bmap_id}", Ansi.LMAGENTA)
+
+    url = f"https://old.ppy.sh/osu/{bmap_id}"
+    async with app.state.services.http_client.get(url) as resp:
+        if resp.status != 200:
+            if 400 <= resp.status < 500:
```
NiceAesth, line 88: i feel like this entire thing could just be simplified into a `resp.ok` check

app/objects/beatmap.py
arily: this is a well known md5 for empty files, just like 200, 500 below the code. You can google it.

app/objects/beatmap.py
arily: I agree, if we are going to eliminate the magic values this should be gone too.

app/objects/beatmap.py
```diff
@@ -57,33 +57,49 @@ async def api_get_beatmaps(**params: Any) -> Optional[list[dict[str, Any]]]:
...
+                stacktrace = app.utils.get_appropriate_stacktrace()
+                await app.state.services.log_strange_occurrence(stacktrace)
+            return False
+        b_beatmap = await resp.read()
+        bytes_md5 = hashlib.md5(b_beatmap).hexdigest()
+        if bytes_md5 in KNOWN_BAD_FILES_MD5:
+            return False
+        # at least this' not a known bad file
+        osu_file_path.write_bytes(b_beatmap)
+        # some callers have type Any | str on "bmap_md5" so I assume it's optional
```
NiceAesth, line 99: please avoid the use of random unnecessary comments

app/objects/beatmap.py
```diff
@@ -57,33 +57,49 @@ async def api_get_beatmaps(**params: Any) -> Optional[list[dict[str, Any]]]:
...
+        if bmap_md5 and bytes_md5 != bmap_md5:
```
NiceAesth, line 100:
```suggestion
        if bmap_md5 is not None and bytes_md5 is not bmap_md5:
```

app/objects/beatmap.py
arily: If we don't use magic values how can we tell a file is bad?

NiceAesth: if you wish to check against an empty file's md5:

- the current example has very poor readability (readability should be independent from comments). Instead, I suggest calculating the md5 for an empty string when the server is started
- store it in something like `EMPTY_FILE_MD5`. why do we need a list of bad hashes? what purpose would this have? feels like this is a bit of scope creep

NiceAesth: if you need an example to think of how this is done you may check regexes which follow the same pattern using `re.compile`

minisbett: Also how would checking if the file is empty help? If you put a whitespace into it it would go through again

app/objects/beatmap.py
```diff
@@ -57,33 +57,48 @@ async def api_get_beatmaps(**params: Any) -> Optional[list[dict[str, Any]]]:
...
+    )
+    if file_md5 == bmap_md5:
+        return True
+    # need to get the file from the osu!api
+    if app.settings.DEBUG:
+        log(f"Doing osu!api (.osu file) request {bmap_id}", Ansi.LMAGENTA)
+
+    url = f"https://old.ppy.sh/osu/{bmap_id}"
+    async with app.state.services.http_client.get(url) as resp:
+        if False == resp.ok():
```
NiceAesth, line 87:
```suggestion
        if not resp.ok:
```

app/objects/beatmap.py
arily: - I originally write
```python
KNOWN_BAD_FILES_MD5 = [
    "d41d8cd98f00b204e9800998ecf8427e" # empty file 
]
```
formatter made it like that.


- An empty hashes is for future proof to add more known bad files.

NiceAesth: yeah my question was WHAT other files could be bad? if you intend to use this for marking random beatmaps as having "bad" files this does not belong in the codebase. if this is intended to be used for multiple "magic" values then this is also going in the wrong direction

app/objects/beatmap.py
```diff
@@ -57,33 +57,48 @@ async def api_get_beatmaps(**params: Any) -> Optional[list[dict[str, Any]]]:
...
+        log(f"Doing osu!api (.osu file) request {bmap_id}", Ansi.LMAGENTA)
+
+    url = f"https://old.ppy.sh/osu/{bmap_id}"
+    async with app.state.services.http_client.get(url) as resp:
+        if not resp.ok:
+            # client error, report this to cmyui
+            stacktrace = app.utils.get_appropriate_stacktrace()
+            await app.state.services.log_strange_occurrence(stacktrace)
+
+        b_beatmap = await resp.read()
```
NiceAesth, line 92: what does b_beatmap mean

app/objects/beatmap.py
arily: > Also how would checking if the file is empty help? If you put a whitespace into it it would go through again

Thus I put a list there. Now the api endpoint returns an empty file, and whatever it returns in the future i am planning to add it to the list.

app/objects/beatmap.py
arily: bytes of beatmap

app/objects/beatmap.py
```diff
@@ -57,33 +57,48 @@ async def api_get_beatmaps(**params: Any) -> Optional[list[dict[str, Any]]]:
...
+        bytes_md5 = hashlib.md5(b_beatmap).hexdigest()
```
NiceAesth, line 93: looking at this now, this has gone in a completely wrong direction. why are you hashing and THEN checking if the response is an empty hash when you could just check if the response is empty without doing anything involving md5?

app/objects/beatmap.py
NiceAesth: https://github.com/osuAkatsuki/bancho.py/pull/428/files#r1167815566 check this

app/objects/beatmap.py
NiceAesth: `bmap_bytes`

app/objects/beatmap.py
NiceAesth: something like
```py
if bmap_bytes.strip()
```
would work for checking if the response is empty

arily: I am concerning about this endpoint returns something other than empty. I can change it to only check for empty files.

minisbett:
Most cursed PR on bpy

app/objects/beatmap.py
NiceAesth: such as? you have yet to provide an example

NiceAesth: that would be a failure on the server's end

arily: I don't have any now but once got html response on `/b/{beatmapId}`

arily:
> Most cursed PR on bpy

well you see it's a draft..

app/objects/beatmap.py
arily: this is removed due to no longer compare against md5s

arily:
I'll reopen another pr without this commits if it's okay

NiceAesth:
> I'll reopen another pr without these commits if it's okay

i mean we can just squash this pr when merging

app/objects/beatmap.py
NiceAesth: that would be solved by checking the content-type header. that happens when cloudflare throws an error page

arily:
I remembered it: when osu is down the endpoints returns cloudflare error pags in HTML. so a check is still needed

NiceAesth:
> I remembered it: when osu is down the endpoints returns cloudflare error pags in HTML. so a check is still needed well res.ok should handle it also so I am not so sure about this now

https://github.com/osuAkatsuki/bancho.py/pull/428#discussion_r1167834535 yes i already said this

app/objects/beatmap.py
```diff
@@ -65,26 +65,38 @@ async def ensure_local_osu_file(
...
+    async with app.state.services.http_client.get(url) as resp:
+        if resp.headers.get("Content-Disposition", "").startswith("attachment"):
+            # bmap_id is not exists.
+            # consider removing following code as it occurs frequently.
+            stacktrace = app.utils.get_appropriate_stacktrace()
+            await app.state.services.log_strange_occurrence(stacktrace)
+            return False
+
+        bmap_bytes = await resp.read()
+        if bmap_bytes.strip():
```
tsunyoku, line 93: ?

app/objects/beatmap.py
```diff
@@ -65,26 +65,38 @@ async def ensure_local_osu_file(
...
+    )
+    if file_md5 == bmap_md5:
+        return True
+    # need to get the file from the osu!api
+    if app.settings.DEBUG:
+        log(f"Doing osu!api (.osu file) request {bmap_id}", Ansi.LMAGENTA)
+
+    url = f"https://old.ppy.sh/osu/{bmap_id}"
+    async with app.state.services.http_client.get(url) as resp:
+        if resp.headers.get("Content-Disposition", "").startswith("attachment"):
```
cmyui, line 85: old pr but curious -- what's this `Content-Disposition` case for? :eyes: is this an osu-specific behavior? or something else?

cmyui:
Going to close this PR for now -- as mentioned in an above comment, might cherry-pick parts of this to #559

app/objects/beatmap.py
arily: With this header browser will save the file to disk with name if provided, instead of trying to show the content.

@NiceAesth told me I should check if the response is valid with this header.

PR #442, opened 2023-05-22
feat: per-mod personal best score on leaderboard

### LINKED ISSUE #237: Leaderboards issues
So, while ago I found 2 issues and I completly forgot to report them here, so first is

**1. Selected Mods Leaderboard**
When looking at selected mods lb, only pb is shown. What does that mean
For example user *a* has 2 plays on Harumachi Clover, One with DT and one No Mod.
User *a* has personal best on No Mod score, when someone, including user *a* tries to look at selected mods lb with DT, he won't see his score because it's not his PB.

**2. PP sorting, but also score at the same time**
So, I don't know if it's on purpose or not but in mania this is very annoying, anyways. This got tested in mania, idk if it works the same in other modes but probably yes.
When user *a* submits score on map *xyz* with 850,000 score and 220 pp for example, his score will be first
Then user *a* submits another score but this time with DT, he managed to get 770k score and 320pp. His DT score will be displayed as first even though it's not PB, same thing happens on HT. Due to mania score cap being 1 milion, it's really annoying and if it's on purpose then it doesn't make sense in mania. Also I think that this may be related to first bug.

g1-1-1:
# Describe your changes

previously, the query was hardcoded to only find the personal best without accounting for the different leaderboards -- now, it works similarly to bancho.

[video](https://streamable.com/02x8dr)

## Related Issues / Projects

closes #237 

## Checklist


app/api/domains/osu.py
```diff
@@ -1301,12 +1301,17 @@ async def get_leaderboard_scores(
...
     elif leaderboard_type == LeaderboardType.Friends:
         query.append("AND s.userid IN :friends")
         params["friends"] = player.friends | {player.id}
+        query.append("AND s.status = 2")
     elif leaderboard_type == LeaderboardType.Country:
         query.append("AND u.country = :country")
         params["country"] = player.geoloc["country"]["acronym"]
+        query.append("AND s.status = 2")
+    else:
+        query.append("AND s.status = 2")
```
hzswdef, line 1314: Why not use one `if` instead of the 3 same lines?

Like this - 
```py
if leaderboard_type != LeaderboardType.Mods:
    query.append("AND s.status = 2")
```

0i8: sure

0i8:
lgtm 👍

7ez:
> lgtm 👍

cus u made it

tsunyoku (changes requested):
this doesn't account for the fact that the user may have multiple submitted scores for a mod combo, you need to add a 1 score per user constraint to the query

g1-1-1:
> this doesn't account for the fact that the user may have multiple submitted scores for a mod combo, you need to add a 1 score per user constraint to the query

i've ran this in my production server for about 2 months now, every case where i've submitted a score with a different mod combination (which would be the best for that combo) seems to work just fine.. if you still think it is a problem i could add what you've requested but it seems kinda redundant

tsunyoku:
it's not redundant, you're just lucky.

g1-1-1:
[how can one be lucky so many times?](https://streamable.com/8ng6mq) i'll look into it regardless

tsunyoku:
just think about the issue i handed you. it needs to be handled no matter how many cases you show me of it seeming to work correctly

kingdom5500:
> [how can one be lucky so many times?](https://streamable.com/8ng6mq) i'll look into it regardless

i am willing to guess (though i will also test to check once i get time) that the client is generous and automatically filters for the best personal score for each person who is in the leaderboard list given by the server. however, our server is limiting the search to 50 rows, several of which may be from the same person. once the client has filtered this out, there will not necessarily be 50 rows remaining, and your top 50 is no longer a top 50. if that is the case upon testing then i'll let you know so you can refine your patch! :D 

i've implemented a slightly different solution on my own server using some SQL inner join mess, but i don't think it should be particularly taxing on performance? that being said, i haven't been able to thoroughly test it on a large user base. the relevant commit is https://github.com/kingdom5500/bancho.py/commit/a2b1cf257ef2a8621a75cd68b5fe3e398557f1fa

tsunyoku:
i'm also generally concerned by some of the choices you made with that query. selecting all non-bests in every case is an unnecessary performance harm, it is only in the mod leaderboard case that we should be considering multiple scores + filtering one per user. we should preferably use `status = 2` for all other leaderboard types, with `status != 0` being kept for mod leaderboards only

kingdom5500:
> yeah, this is what is happening. didn't state it explicitly as i wanted them to think about the problem, but it's okay.

i know :D it's just that it had been a week without any updates to the PR or the discussion, so i figured i might give some pointers.

> your solution works, but it's not very comfortably readable thinking (namely the `query.insert` choice)

i completely agree. in fact, the `query.append` is totally redundant alongside the `query.insert`, so the append shouldn't even be there. even still, it would be more readable if i had just used a `COALESCE` of some sorts instead. thanks for reminding me to think about this, i had meant to make it more readable before making the commit but was too tired to bother.

> i'm also generally concerned by some of the choices you made with that query. selecting all submitted scores in every case is an unnecessary performance harm, it is only in the mod leaderboard case that we should be considering multiple scores + filtering one per user. we should preferably use status = 2 for all other leaderboard types, with status != 0 being kept for mod leaderboards only. the extra join is also an unnecessary performance harm outside of the mod leaderboard scenario

in my own server, this behaviour is intentional since it mostly matches what you get on the ppy bancho leaderboards, though that's just based on my own and my players' preferences. i didn't intend for my patch to be a drop-in solution by any means! instead, refer to the discussion in #237, where there hasn't really been a firm decision on what the general behaviour should be for bancho.py.

regardless of this, i understand that the query may not be very optimal! i'm just not very well-practised in the way of sql just yet. i'm working on it though hehe

tsunyoku:
heh that's no problem! just wanted to give some feedback :p

i'll let this pr sit for a bit given the discussion and see if @g1-1-1 decides to make any changes. if not, i'll close it and keep it as an open issue for now

7ez:
jesus christ yall are shitting on jupiter rn

g1-1-1:
i'll work on this in a few, you can limit 1 per user using raw sql but it is tricky to pull off

cmyui:
I'll close this PR for now -- I've created an issue to track this feature request: #560

PR #449, opened 2023-06-19
feat: use `smart_open` for data endpoint variety

stevefan1999-personal:
# Describe your changes

## Checklist
- [ ] The changes pass pre-commit checks (`make lint`)


tsunyoku:
if i'm being completely real - i do not like this approach to trying to support other means of storing files (s3, etc.)

i'd rather build adapters for these different types and have them usable via config options. something like s3 especially would be easy, we have built these adapters and use them on production @ akatsuki in numerous different places.

patching just feels messy. and it makes code unintuitive imo. open to other opinions but i really don't think a botch patch is the method here

PR #450, opened 2023-06-19
feat: add helm chart

stevefan1999-personal:
# Describe your changes

## Checklist
- [ ] Get official Docker repo and image
- [ ] Find a way to integrate [go-migrate](https://github.com/golang-migrate/migrate) because it fits well with [init container](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/) (which is done by Gitea Helm chart)
- [ ] Fix all the shit template
- [ ] Make sure that the chart values README are generated (that means documenting the values)
- [ ] Deploy a test pipeline using [vcluster](https://www.vcluster.com/)
- [ ] Get a Github page to publish the charts (that means adding the [chart-releaser action](https://github.com/marketplace/actions/helm-chart-releaser))

### DIFF
--- a/charts/banchopy/.gitignore
+++ b/charts/banchopy/.gitignore
@@ -1,0 +1,2 @@
+charts/**.tgz
+.template
--- a/charts/banchopy/.helmignore
+++ b/charts/banchopy/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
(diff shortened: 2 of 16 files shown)


cmyui:
Hey, this is neat the work you've done here but I don't think this is particularly useful for most users of b.py - at least in it's current state - I think it would make most sense as a repo you build as a fork of this project.

I think there are quite a few roadblocks before b.py will be worthy of throwing on k8s - notably making it fully stateless

I'd be open to making an official helm chart in the future when the time is appropriate & the project is in the right state for it - but not sure if it should even live in this repo (it's already quite bloated imo)

Ty for the PR!

cmyui:
> I think there are quite a few roadblocks before b.py will be worthy of throwing on k8s - notably making it fully stateless

made a ticket for this for @tsunyoku and i
- https://github.com/osuAkatsuki/bancho.py/issues/506

PR #459, opened 2023-07-18
fix: get_replay properly appending headers to replays

### LINKED ISSUE #458: bug: v1 api endpoint /get_replay appears to be incomplete
### Describe the bug

replays downloaded from `v1/get_replay` cannot be processed by the client, because the replay headers are missing regardless of the `include_headers` parameter. presumably, `include_headers=True` should return a functioning `.osr` file, but the server code [seems to process it the other way around](https://github.com/osuAkatsuki/bancho.py/blob/65fabad0849c6eda3252f81ea11aa1c472cc7e28/app/api/v1/api.py#L740).

even with this in mind, specifying `include_headers=False` still provides the same incomplete replay file because the `replay_data` variable which contains the headers is never actually affixed to the raw replay file before it is sent off.

### To Reproduce

1. download the .osr file from https://api.skrungly.com/v1/get_replay?id=1447&include_headers=True (also ensure you have [the map](https://osu.ppy.sh/beatmapsets/677687#osu/1490588))
2. import into your game
3. stonks!

image

### Expected behavior

when specifying `include_headers=True`, the `.osr` file should include headers. 

### bancho.py Version

latest

### Python Version

3.9.X (Default)

### Relevant log output

_No response_

### Additional context

i understand that the v1 api is soon to be deprecated, but to my knowledge it is currently the only way to receive certain data (such as whole replay files) from the server, and so it should at least be functional in the meantime.

g1-1-1:
# Describe your changes
uses Response() with bytes instead of FileResponse() with a path, make include_header default to true (since there are not many usecases where we don't need to add headers to replays)

## Related Issues / Projects
should fix #458

## Checklist


app/api/v1/api.py
```diff
@@ -789,12 +782,14 @@ async def api_get_replay(
...
     ).hexdigest()
-
     # create a buffer to construct the replay output
     replay_data = bytearray()
-
     # pack first section of headers.
-    replay_data += struct.pack("<Bi", row["mode"], 20200207)  # TODO: osuver
+    replay_data += struct.pack(
+        "<Bi",
+        as_vanilla(row["mode"]),
```
kingdom5500, line 790: this doesn't work, should be `GameMode(row["mode"]).as_vanilla,`

app/api/v1/api.py
```diff
@@ -812,23 +807,18 @@ async def api_get_replay(
...
     # pack additional info buffer.
     replay_data += struct.pack("<q", score_id)
-
     # NOTE: target practice sends extra mods, but
     # can't submit scores so should not be a problem.
-
     # stream data back to the client
-    return FileResponse(
-        path=REPLAYS_PATH / f"{score_id}.osr",
+    return Response(
```
kingdom5500, line 820: you need to import `Response`

app/api/v1/api.py
g1-1-1: huh?

g1-1-1: oh wow sorry, i use a different version of the api which doesn't require this

g1-1-1:
works fine now :) thanks for the second pass

kingdom5500:
can confirm, this patch now works correctly.

kingdom5500 (approved):

NiceAesth:
It looks fine to me for now, the logic for this needs to be refactored in the future either way. Will see what others say about this.

app/api/v1/api.py
```diff
@@ -722,35 +722,31 @@ async def api_get_score_info(
 @router.get("/get_replay")
 async def api_get_replay(
     score_id: int = Query(..., alias="id", ge=0, le=9_223_372_036_854_775_807),
-    include_headers: bool = False,
+    include_headers: bool = True,
```
tsunyoku, line 725: is there a particular reason this default changed? i'm not arguing for or against the default being True, but rather the fact it is a change in behaviour

app/api/v1/api.py
7ez: > this doesn't work, should be `GameMode(row["mode"]).as_vanilla,`

why not `row["mode"] % 4` instead?

NiceAesth: then you would have duplicated logic and if by any chance gamemodes get changed you would have random places to change this

app/api/v1/api.py
g1-1-1: we don't, currently, need to utilise include_headers enough to explicitly declare it True every time, since in 99/100 cases when we use this endpoint, it is a frontend grabbing a replay

it simplifies the use of this endpoint, and as a result, users won't have to explicitly declare include_headers=True, since it will already be default for True

PR #425, opened 2023-04-10
add separate config for docker and manual install

cfgexe:
# Describe your changes
Add seperate configs for docker and manual
## Checklist
- [ ] The changes pass pre-commit checks (`make lint`)
- [ ] The changes follow coding style
i wonder how fucked this is

PR #463, opened 2023-07-19
chore: (Docker) Use env variables in the nginx config for port and domain

7mochi:
# Describe your changes

## Checklist


docker-compose.yml
```diff
@@ -57,13 +57,15 @@ services:
     depends_on:
       - bancho
     volumes:
-      - ./ext/nginx.conf:/etc/nginx/conf.d/default.conf:ro
+      - ./ext/docker-nginx.conf.example:/etc/nginx/templates/nginx.conf.template:ro
```
NiceAesth, line 60: very odd choice to use an "example" file in here, when the .env file is with the normal name

.github/docs/wiki/Setting-up-(Docker).md
```diff
@@ -39,17 +39,6 @@ cp .env.example .env
 nano .env
 ```
 
-## configuring a reverse proxy (we'll use nginx)
```
NiceAesth, line 42: these changes should also appear in the translations

NiceAesth (changes requested):

docker-compose.yml
7mochi: the filename was changed in #425 so I used the file as is

NiceAesth: That PR is a bit problematic

# create a configuration file from the sample provided
cp .env.example .env

this is now incorrect in both setup methods (+ translations)

also, the example cannot remain in the compose. i propose adding instructions to do something similar to .env, or to just rename the file and move it in `docker/` (could be the better solution)

NiceAesth (approved):

PR #477, opened 2023-08-11
Set an absolute limit of 1 day on beatmaps cache; generalize some code

cmyui:
# Describe your changes

## Checklist

### DIFF
--- a/app/objects/beatmap.py
+++ b/app/objects/beatmap.py
@@ -588,35 +588,18 @@ class BeatmapSet:
         """The online url for this beatmap set."""
         return f"https://osu.{app.settings.DOMAIN}/beatmapsets/{self.id}"
 
-    def all_officially_ranked_or_approved_or_frozen(self) -> bool:
-        """Whether all the maps in the set are
-        ranked or approved on official servers."""
-        return all(
-            # ranked status has been edited on bancho.py
-            bmap.frozen or
-            # ranked status is ranked or approved on bancho
-            bmap.status in (RankedStatus.Ranked, RankedStatus.Approved)
-            for bmap in self.maps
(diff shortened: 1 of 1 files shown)


cmyui:
Not yet tested

NiceAesth:
Are you still looking into this?

cmyui:
> Are you still looking into this?

Can't find time to test it :( but this is 99% a fix for the problem where people have beatmaps they cannot update

NiceAesth:
Any repro steps in case I have the time to test it?

app/objects/beatmap.py
```diff
@@ -588,35 +588,18 @@ def url(self) -> str:  # same as above, just no beatmap id
...
-            bmap.frozen or
-            # ranked status is loved on bancho
-            bmap.status == RankedStatus.Loved
-            for bmap in self.maps
+    def any_beatmaps_have_official_leaderboards(self) -> bool:
+        """Whether all the maps in the set have leaderboards on official servers."""
+        leaderboard_having_statuses = (
+            RankedStatus.Loved,
+            RankedStatus.Ranked,
+            RankedStatus.Approved,
```
tsunyoku, line 596:
```suggestion
            RankedStatus.Approved,
            RankedStatus.Qualified,
```

tsunyoku: qualified queue is generally quite slow on updates (~7 days after qualification) so it's safe & likely more efficient to treat it the same as the rest

cmyui: i think they're different enough to warrant different cases to avoid the case of incorrectly high confidence near the end of the 7d window

cmyui:
going to get this in since it's quick fix to some cache issues people are having

PR #479, opened 2023-08-11
fix: prevent crash when leaving tournament match channel

KhooHaoYit:
# Describe your changes
This PR fixes crashes caused by the tournament client when removing attempting to remove players who are not in the match.

## Checklist

Im having issue running `make lint` in my system from `setuptools` complaining and python wasn't my main language soo it would be nice to know how to run `make lint` properly

### DIFF
--- a/app/api/domains/cho.py
+++ b/app/api/domains/cho.py
@@ -2002,7 +2002,8 @@ class TourneyMatchLeaveChannel(BasePacket):
 
         # attempt to join match chan
         player.leave_channel(match.chat)
-        match.tourney_clients.remove(player.id)
+        if player.id in match.tourney_clients:
+            match.tourney_clients.remove(player.id)
 
 
 @register(ClientPackets.FRIEND_ADD)
--- a/app/api/domains/cho.py
+++ b/app/api/domains/cho.py
@@ -1997,13 +1997,12 @@ class TourneyMatchLeaveChannel(BasePacket):
             return  # insufficient privs
 
         match = app.state.sessions.matches[self.match_id]
-        if not match:
+        if not (match and player.id in match.tourney_clients):
             return  # match not found
 
         # attempt to join match chan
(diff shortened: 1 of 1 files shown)


NiceAesth:
As a note to the other maintainers, I do not think that this is the right direction to go about this and should at most be considered a hotfix. Instead it should be looked into as to why the situation happens in the first place.

cmyui:
hmm let me look into the root cause of the error

cmyui:
Ok I did some more thinking about this -- this happens because osu! servers do not have any mechanism to inform tourney clients when a match that they're spectating has been disbanded. The tourney client continues to spec as if nothing happened.

This gets ultra weird with bancho.py's match id incrementing behaviour -- which allocates a static 64 match slots, and attempts to re-use those ids.

The sequence of events leading to this error is:
1. Tourney match is created; tourney client starts spectating the match.
2. Tourney match is disbanded; client continues spectating the players.
3. A new match is created, with the same id as the original tourney match.
4. The tourney client switches to spectating _any other_ match, causing it to attempt leaving the old match's channel (by id), using the `TOURNAMENT_LEAVE_MATCH_CHANNEL` packet (#109).

In this case, the server will resolve the wrong match id, allow past the `if not match:` check, and attempt to remove them from the `match.tourney_clients`, which will fail.

I think the BEST fix here is to migrate towards monotonically incrementing (universally-unique) match ids; but that's a decently sized refactor. I think this PR is a _good enough_ interim solution, so we can merge it.

cmyui:
I'm going to make one small adjustment, to move the `player.id in match.tourney_clients:` check to happen before the `player.leave_channel(match.chan)` -- to have it alongside the `if not match`.

cmyui (approved):

cmyui:
Ty for the PR @KhooHaoYit!! 🎉 And sorry for the large delay 🤣

PR #486, opened 2023-09-11
refactor: comments repository

7mochi:
# Describe your changes

## Checklist

### DIFF
--- a/app/repositories/comments.py
+++ b/app/repositories/comments.py
@@ -1,0 +1,213 @@
+from __future__ import annotations
+
+import textwrap
+from typing import Any
+from typing import Optional
+
+import app.state.services
+
+# +-----------------+-----------------------------+------+-----+---------+----------------+
+# | Field           | Type                        | Null | Key | Default | Extra          |
+# +-----------------+-----------------------------+------+-----+---------+----------------+
+# | id              | int                         | NO   | PRI | NULL    | auto_increment |
+# | target_id       | int                         | NO   |     | NULL    |                |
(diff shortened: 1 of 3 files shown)


app/repositories/comments.py
```diff
@@ -0,0 +1,238 @@
...
+    """\
+        id, target_id, target_type, userid, time, comment, colour
+    """,
+)
+
+
+class Comment(TypedDict):
+    id: int
+    target_id: int
+    target_type: str
```
cmyui, line 34: can/should this be an enum? it is in db

app/repositories/comments.py
```diff
@@ -0,0 +1,238 @@
...
+        "score_id": score_id,
+        "map_set_id": map_set_id,
+        "map_id": map_id,
+    }
+
+    comments = await app.state.services.database.fetch_all(query, params)
+    return cast(list[Comment], [dict(c._mapping) for c in comments])
+
+
+async def update(
```
cmyui, line 172: i don't think we need update/delete for comments - should be create&read only i think

app/repositories/comments.py
```diff
@@ -0,0 +1,113 @@
...
+        "id": rec_id,
+    }
+    _comment = await app.state.services.database.fetch_one(query, params)
+
+    assert _comment is not None
+    return cast(Comment, dict(_comment._mapping))
+
+
+class CommentWithUserPrivileges(Comment):
+    priv: int
```
cmyui, line 84: cursed abstraction but probably better than O(n) db calls

cmyui:
tested on local and is working

cmyui (approved):

PR #489, opened 2023-09-12
refactor: ingame_logins repository

7mochi:
# Describe your changes

## Checklist

### DIFF
--- a/app/repositories/ingame_logins.py
+++ b/app/repositories/ingame_logins.py
@@ -1,0 +1,232 @@
+from __future__ import annotations
+
+import textwrap
+from typing import cast
+from typing import TypedDict
+
+import app.state.services
+from app._typing import _UnsetSentinel
+from app._typing import UNSET
+
+# +--------------+------------------------+------+-----+---------+-------+
+# | Field        | Type                   | Null | Key | Default | Extra |
+# +--------------+------------------------+------+-----+---------+-------+
+# | id           | int                    | NO   | PRI | NULL    |       |
+# | userid       | int                    | NO   |     | NULL    |       |
+# | ip           | varchar(45)            | NO   |     | NULL    |       |
+# | osu_ver      | date                   | NO   |     | NULL    |       |
+# | osu_stream   | varchar(11)            | NO   |     | NULL    |       |
+# | datetime     | datetime               | NO   |     | NULL    |       |
+# +--------------+------------------------+------+-----+---------+-------+
+
+READ_PARAMS = textwrap.dedent(
+    """\
+        id, userid, ip, osu_ver, osu_stream, datetime
+    """,
+)
+
+
+class InGameLogin(TypedDict):
+    id: int
+    userid: str
+    ip: str
+    osu_ver: str
+    osu_stream: str
+    datetime: datetime
+
+
+class InGameLoginUpdateFields(TypedDict, total=False):
+    userid: str
+    ip: str
+    osu_ver: str
+    osu_stream: str
+
+
+async def create(
+    user_id: int,
+    ip: str,
+    osu_ver: str,
+    osu_stream: str,
+) -> InGameLogin:
+    """Create a new login entry in the database."""
+    query = f"""\
+        INSERT INTO ingame_logins (userid, ip, osu_ver, osu_stream, datetime)
+             VALUES (:userid, :ip, :osu_ver, :osu_stream, NOW())
+    """
+    params = {
+        "userid": user_id,
+        "ip": ip,
+        "osu_ver": osu_ver,
+        "osu_stream": osu_stream,
+    }
+    rec_id = await app.state.services.database.execute(query, params)
+
+    query = f"""\
+        SELECT {READ_PARAMS}
+          FROM ingame_logins
+         WHERE id = :id
+    """
+    params = {
+        "id": rec_id,
+    }
+    ingame_login = await app.state.services.database.fetch_one(query, params)
+
+    assert ingame_login is not None
+    return cast(InGameLogin, ingame_login)
+
+
+async def fetch_one(
+    id: int | None = None,
+    user_id: int | None = None,
+    ip: str | None = None,
+    osu_ver: str | None = None,
+    osu_stream: str | None = None,
+) -> InGameLogin | None:
+    """Fetch a login entry from the database."""
+    if (
+        id is None
+        and user_id is None
+        and ip is None
+        and osu_ver is None
+        and osu_stream is None
+    ):
+        raise ValueError("Must provide at least one parameter.")
+
+    query = f"""\
+        SELECT {READ_PARAMS}
+          FROM ingame_logins
+         WHERE id = COALESCE(:id, id)
+           AND userid = COALESCE(:userid, userid)
+           AND ip = COALESCE(:ip, ip)
+           AND osu_ver = COALESCE(:osu_ver, osu_ver)
+           AND osu_stream = COALESCE(:osu_stream, osu_stream)
+    """
+    params = {
+        "id": id,
+        "userid": user_id,
+        "ip": ip,
+        "osu_ver": osu_ver,
+        "osu_stream": osu_stream,
+    }
+    ingame_login = await app.state.services.database.fetch_one(query, params)
+
+    return cast(InGameLogin, ingame_login) if ingame_login is not None else None
+
+
+async def fetch_count(
+    user_id: int | None = None,
+    ip: str | None = None,
+) -> int:
+    """Fetch the number of logins in the database."""
+    query = """\
+        SELECT COUNT(*) AS count
+          FROM ingame_logins
+        WHERE userid = COALESCE(:userid, userid)
+          AND ip = COALESCE(:ip, ip)
+    """
+    params = {
+        "userid": user_id,
+        "ip": ip,
+    }
+    rec = await app.state.services.database.fetch_one(query, params)
+    assert rec is not None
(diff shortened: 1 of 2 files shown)


app/repositories/ingame_logins.py
```diff
@@ -0,0 +1,213 @@
...
+# +--------------+------------------------+------+-----+---------+-------+
+
+READ_PARAMS = textwrap.dedent(
+    """\
+        id, userid, ip, osu_ver, osu_stream, datetime
+    """,
+)
+
+
+async def create(userid: int, ip: str, osu_ver: str, osu_stream: str) -> dict[str, Any]:
```
cmyui, line 27: nit: use user_id in the code, and only map to `userid` when interacting with database (at repository layer)

app/repositories/ingame_logins.py
```diff
@@ -0,0 +1,213 @@
...
+    assert rec is not None
+    return dict(rec)
+
+
+async def fetch_one(
+    id: Optional[int] = None,
+    userid: Optional[int] = None,
+    ip: Optional[str] = None,
+    osu_ver: Optional[str] = None,
+    osu_stream: Optional[str] = None,
```
cmyui, line 59: only params which uniquely identify an ingame login should be accepted into fetch_one()

params which identify multiple should be used with fetch_many()

cmyui: (similarly, fetch_many should not have params which uniquely identify a single record - which seems correctly done in this pr)

app/repositories/ingame_logins.py
```diff
@@ -0,0 +1,213 @@
...
+    recs = await app.state.services.database.fetch_all(query, params)
+    return [dict(rec) for rec in recs]
+
+
+async def update(
+    id: int,
+    userid: Optional[int] = None,
+    ip: Optional[str] = None,
+    osu_ver: Optional[str] = None,
+    osu_stream: Optional[str] = None,
```
cmyui, line 159: do we have an `Unset` pattern in bancho.py repositories? I can't remember if I ever implemented that here

if not we might want to figure it out sooner than later so that we can differentiate between nulls and unset values

7mochi: we don't

NiceAesth:
Will you apply the unset suggestion to all of the currently open PRs or will that be a following PR?

Would prefer if  it was multiple palletable prs

7mochi:
Unset suggestion applied

app/repositories/ingame_logins.py
```diff
@@ -0,0 +1,234 @@
...
+# +--------------+------------------------+------+-----+---------+-------+
+
+READ_PARAMS = textwrap.dedent(
+    """\
+        id, userid, ip, osu_ver, osu_stream, datetime
+    """,
+)
+
+
+class InGameLogin(TypedDict):
```
cmyui, line 31: nit: `IngameLogin`

`InGameLogin` would be for `in_game_logins` in db

app/repositories/ingame_logins.py
```diff
@@ -0,0 +1,233 @@
...
+          FROM ingame_logins
+         WHERE id = :id
+    """
+    params = {
+        "id": rec_id,
+    }
+    ingame_login = await app.state.services.database.fetch_one(query, params)
+
+    assert ingame_login is not None
+    return cast(IngameLogin, ingame_login)
```
cmyui, line 76:
```suggestion
    return cast(IngameLogin, dict(ingame_login._mapping))
```

cmyui: pls do for all cases when returning from repos - we adjusted all other cases in #501

app/repositories/ingame_logins.py
```diff
@@ -0,0 +1,233 @@
...
+        WHERE userid = COALESCE(:userid, userid)
+          AND ip = COALESCE(:ip, ip)
+    """
+    params = {
+        "userid": user_id,
+        "ip": ip,
+    }
+    rec = await app.state.services.database.fetch_one(query, params)
+    assert rec is not None
+    return rec["count"]
```
cmyui, line 134: rec._mapping["count"] is fine here - no need to make a copy because we do not need to return a mutable copy

app/repositories/ingame_logins.py
cmyui: ingame_login by itself will be a Record instance from aiomysql

ingame_login._mapping gives a Mapping[str, Any], but it's readonly

it's useful to return an updatable model imo, so i typically copy into a new dict

app/repositories/ingame_logins.py
```diff
@@ -0,0 +1,233 @@
...
+        update_fields["osu_ver"] = osu_ver
+    if not isinstance(osu_stream, _UnsetSentinel):
+        update_fields["osu_stream"] = osu_stream
+
+    query = f"""\
+        UPDATE ingame_logins
+           SET {",".join(f"{k} = :{k}" for k in update_fields)}
+         WHERE id = :id
+    """
+    values = {"id": id} | update_fields
```
cmyui, line 196: should probably be "params" to match rest of code, might switch to "values" sometime to match lower level api so i don't really mind

app/repositories/ingame_logins.py
```diff
@@ -0,0 +1,233 @@
...
+
+    query = """\
+        DELETE FROM ingame_logins
+              WHERE id = :id
+    """
+    params = {
+        "id": id,
+    }
+    await app.state.services.database.execute(query, params)
+    return cast(IngameLogin, ingame_login) if ingame_login is not None else None
```
cmyui, line 233: can be an assert right? i think it cannot be null here

app/repositories/ingame_logins.py
```diff
@@ -0,0 +1,233 @@
...
+    assert ingame_login is not None
+    return cast(IngameLogin, ingame_login)
+
+
+async def fetch_one(
+    id: int | None = None,
+    user_id: int | None = None,
+    ip: str | None = None,
+    osu_ver: str | None = None,
+    osu_stream: str | None = None,
```
cmyui, line 84: this should only have `id` - all the others will return many results, not one

app/repositories/ingame_logins.py
```diff
@@ -0,0 +1,233 @@
...
+           OFFSET :offset
+        """
+        params["limit"] = page_size
+        params["offset"] = (page - 1) * page_size
+
+    ingame_logins = await app.state.services.database.fetch_all(query, params)
+    return cast(list[IngameLogin], ingame_logins)
+
+
+async def update(
```
cmyui, line 173: i think ingame_logins should be only creatable and not updatable - it's a historical record

app/repositories/ingame_logins.py
```diff
@@ -0,0 +1,233 @@
...
+        WHERE id = :id
+    """
+    params = {
+        "id": id,
+    }
+    ingame_login = await app.state.services.database.fetch_one(query, params)
+    return cast(IngameLogin, ingame_login) if ingame_login is not None else None
+
+
+async def delete(id: int) -> IngameLogin | None:
```
cmyui, line 211: also doesn't need delete - shouldn't be deleting these ever

app/repositories/ingame_logins.py
```diff
@@ -0,0 +1,152 @@
...
+        "id": rec_id,
+    }
+    ingame_login = await app.state.services.database.fetch_one(query, params)
+
+    assert ingame_login is not None
+    return cast(IngameLogin, dict(ingame_login._mapping))
+
+
+async def fetch_one(
+    id: int | None = None,
```
cmyui, line 80: no need to accept null case

app/repositories/ingame_logins.py
```diff
@@ -0,0 +1,152 @@
...
+) -> IngameLogin | None:
+    """Fetch a login entry from the database."""
+    if id is None:
+        raise ValueError("Must provide at least one parameter.")
```
cmyui, line 84: or check here

app/repositories/ingame_logins.py
```diff
@@ -0,0 +1,152 @@
...
+
+    query = f"""\
+        SELECT {READ_PARAMS}
+          FROM ingame_logins
+         WHERE id = COALESCE(:id, id)
```
cmyui, line 89: or use coalesce

cmyui:
i think the rest is good

app/repositories/ingame_logins.py
```diff
@@ -0,0 +1,149 @@
...
+    ip: str,
+    osu_ver: str,
+    osu_stream: str,
+) -> IngameLogin:
+    """Create a new login entry in the database."""
+    query = f"""\
+        INSERT INTO ingame_logins (userid, ip, osu_ver, osu_stream, datetime)
+             VALUES (:userid, :ip, :osu_ver, :osu_stream, NOW())
+    """
+    params = {
```
cmyui, line 57: ah we should also set `dict[str, Any]` as the type on each declaration of `params` because the default behaviour will be to take the types of the initial definition, which will be more specific than dict[str, Any]

cmyui: but this should only be set on the initial declaration of params - if it's redefined in the same function ,it should not have the type annotation

mypy will point this out

cmyui: u can see it in the other repos from master after #501

cmyui: pls apply to all prs as well

cmyui (approved):

PR #496, opened 2023-09-21
RFC: tear out reverse proxy specific elements

NiceAesth:
# Describe your changes

Removed nginx from docker configs, running a reverse proxy inside of a docker configuration has always been questionable to me. Additionally, depending on a specific nginx configuration is a bad idea (so is providing production configs and assuming everyone will use them).

To-do:
- [x] Test Caddy config
- [ ] Simplify nginx config (potentially)
- [ ] Change code in `services.py` to no longer rely on nginx
- [ ] Change lookup strategy in `services.py` to not fail if `X-Real-IP` is missing
- [ ] Update documentation


tsunyoku:
idea of RFC i agree with, but yeah needs to make sure that bancho.py is well equipped to handle essentially any reverse proxy we can think of. also would require a somewhat large README adjustment ofc because it will be more of a "expose this in your own way" than "here is our nginx config, and it is gospel"

docker/nginx-with-geoip2/Dockerfile
cmyui, line 1: we should use cloudflare for geoip

tsunyoku: yes if available ^
but should not be assumed behaviour

PR #508, opened 2023-09-24
Tests running in CI & score submission integration test

cmyui:
# Describe your changes

## Checklist


cmyui:
@tsunyoku do u know why i'm getting this mypy error?

> tests/integration/domains/osu_test.py:14: error: Function is missing a return type annotation  [no-untyped-def]

by my understanding `mypy-tests.*` should be covering this - and it's covering those under `tests/unit`..

cmyui:
I don't think I'll merge this until it is backed by a temporary db.

tsunyoku:
> @tsunyoku do u know why i'm getting [this mypy error](https://github.com/osuAkatsuki/bancho.py/actions/runs/6295660521/job/17089276157)?
> 
> 
> 
> > tests/integration/domains/osu_test.py:14: error: Function is missing a return type annotation  [no-untyped-def]
> 
> 
> 
> by my understanding `mypy-tests.*` should be covering this - and it's covering those under `tests/unit`..
> 
> 
> 
> https://github.com/osuAkatsuki/bancho.py/blob/58ef2b5bc79f3cdcb4bb66cab55465c95c5edfac/mypy.ini#L5-L7

i'm not sure how this plays with module paths over regular packages, i never really excluded via path with mypy before

cmyui:
runs now

image

tests/integration/domains/osu_test.py
```diff
@@ -0,0 +1,208 @@
...
+        },
+        files={
+            # simulate replay data
+            "score": b"12345"
+            * 100,
+        },
+    )
+
+    # ASSERT
+    assert response.status_code == status.HTTP_200_OK
```
tsunyoku, line 208: we should actually check the response doesn't start with "error: " here, because we respond with 200 even on error cases

tsunyoku:
would also be nice to add a load leaderboard step before score sub to better test the real flow that would happen? would also be better for JITing the beatmap

tests/integration/domains/osu_test.py
cmyui: yes i still have a lot of things to assert - would be perhaps good to install some function call spies as well to measure what's being performed

tsunyoku: yep, we can probably assert the entire chart response even

cmyui: i will try to get this running in ci & merged so we can iterate on it in future prs

cmyui:
tests are passing in CI now

main.py
```diff
@@ -110,6 +110,7 @@ def main(argv: Sequence[str]) -> int:
         ) from None
 
     # run the server indefinitely
+    # TODO: gracefully handle SIGTERM or move to gunicorn
```
cmyui, line 113:
```suggestion
```

app/api/domains/osu.py
```diff
@@ -1083,12 +1058,7 @@ async def osuSubmitModularSelector(
     """ score submission charts """
 
     # charts are only displayed for passes vanilla gamemodes.
-    if not score.passed or score.mode not in (
-        GameMode.VANILLA_OSU,
-        GameMode.VANILLA_TAIKO,
-        GameMode.VANILLA_CATCH,
-        GameMode.VANILLA_MANIA,
-    ):
+    if not score.passed:  # TODO: check if this is correct
```
cmyui, line 1063: Behaviour change here -- but I think this check was not useful/overkill logic

tests/integration/domains/osu_test.py
```diff
@@ -0,0 +1,212 @@
...
+from datetime import datetime
+from uuid import UUID
+
+from fastapi import status
+from httpx import AsyncClient
+
+from app import encryption
+
+
+async def test_score_submission(http_client: AsyncClient) -> None:
```
cmyui, line 14: This is currently making real 3rd party (e.g. osu! api) calls in CI.

Probably not ideal -- would prefer to mock out 3rd party calls to reduce test flakiness & abuse, though it makes for quite the good integration test.

cmyui: Not 100% sure how I would ideally resolve this -- perhaps we could have some test data in the repo, such as .osu files, etc. such that we can continue to itest the majority of the code flows, but without the 3rd party network calls.

cmyui, line 14: This is currently very low-level in implementation. I think I'll defer the extraction/abstraction work to a future PR, as it will be more clear what should be extracted once there are a couple of tests.

cmyui: Nvm -- resolving it in this PR here: https://github.com/osuAkatsuki/bancho.py/pull/508/commits/cc07c5d00147156449bf3e6e6cec8ecad8c7105d

PR #511, opened 2023-09-24
fix: remove any linux specific implementations

### LINKED ISSUE #502: fix: remove linux-specific code
Some examples:
https://github.com/osuAkatsuki/bancho.py/blob/92a4e1c58e9e1509009f4fe3ef6d99f8811affc7/app/utils.py#L411-L422
https://github.com/osuAkatsuki/bancho.py/blob/92a4e1c58e9e1509009f4fe3ef6d99f8811affc7/app/api/init_api.py#L122-L126
https://github.com/osuAkatsuki/bancho.py/blob/92a4e1c58e9e1509009f4fe3ef6d99f8811affc7/app/commands.py#L1220

tsunyoku:
# Describe your changes

Removes any code that was written specifically for Linux in mind in preparation for Windows support.

Note that uvloop still needs to be replaced with winloop before Windows will properly work.

## Related Issues / Projects

Resolves #502

## Checklist

### DIFF
--- a/app/commands.py
+++ b/app/commands.py
@@ -9,7 +9,6 @@ import signal
 import time
 import traceback
 import uuid
-from collections import Counter
 from collections.abc import Awaitable
 from collections.abc import Callable
 from collections.abc import Mapping
@@ -27,6 +26,7 @@ from typing import TYPE_CHECKING
 from typing import TypedDict
 from urllib.parse import urlparse
 
+import cpuinfo
 import psutil
 import timeago
 from pytimeparse.timeparse import timeparse
@@ -1227,18 +1227,13 @@ async def server(ctx: Context) -> str | None:
     uptime = int(time.time() - proc.create_time())
 
     # get info about our cpu
-    with open("/proc/cpuinfo") as f:
-        header = "model name\t: "
-        trailer = "\n"
-
-        model_names = Counter(
-            line[len(header) : -len(trailer)]
-            for line in f.readlines()
-            if line.startswith("model name")
-        )
+    cpu_info = cpuinfo.get_cpu_info()
 
     # list of all cpus installed with thread count
-    cpus_info = " | ".join(f"{v}x {k}" for k, v in model_names.most_common())
+    thread_count = cpu_info["count"]
+    cpu_name = cpu_info["brand_raw"]
+
+    cpu_info_str = f"{thread_count}x {cpu_name}"
 
     # get system-wide ram usage
     sys_ram = psutil.virtual_memory()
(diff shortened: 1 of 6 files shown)


tsunyoku:
this is how `!server`'s cpu info now looks for those interested:
image

NiceAesth:
Copying from discord

We probably have to reconsider whether a single line in a command to display cpu info warrants a whole dependency because we have been quite heavy handed with them

(Or if that package provides anything useful in the future)

tsunyoku:
to note on that one there isn't any worth on building functions to display the cpu info ourselves because it's a lot of silly parsing and the dependency we use is already very minimal

so the argument is we either keep the dependency and display cpu info, or we don't display cpu info at all

app/commands.py
```diff
@@ -1227,18 +1227,13 @@ async def server(ctx: Context) -> str | None:
...
-            if line.startswith("model name")
-        )
+    cpu_info = cpuinfo.get_cpu_info()
 
     # list of all cpus installed with thread count
-    cpus_info = " | ".join(f"{v}x {k}" for k, v in model_names.most_common())
+    thread_count = cpu_info["count"]
+    cpu_name = cpu_info["brand_raw"]
+
+    cpu_info_str = f"{thread_count}x {cpu_name}"
```
cmyui, line 1236: can we use shutil?

tsunyoku: if you're prepared to write parsers for the n amount of different os specific commands for this be my guest

it's just not worth it 😆

cmyui: oops sorry i meant psutil, not shutil..

i think it has multiplatform support and we already have it installed

tsunyoku: it can be used to provide usage information but not threads / name

if you want to display cpu usage instead then sure, i just wanted to mimic previous behaviour

cmyui: we can get thread count with https://psutil.readthedocs.io/en/latest/#psutil.cpu_count

i'm not sure how to do cpu name though.. we can get architecture & processor using https://docs.python.org/3/library/platform.html but idk how py-cpuinfo is doing the full model name

app/commands.py
```diff
@@ -1275,7 +1270,7 @@ async def server(ctx: Context) -> str | None:
     return "\n".join(
         (
             f"{build_str} | uptime: {seconds_readable(uptime)}",
-            f"cpu(s): {cpus_info}",
+            f"cpu(s): {cpu_info_str}",
```
cmyui, line 1273:
```suggestion
            f"cpu: {cpu_info_str}",
```

cmyui (approved):

PR #512, opened 2023-09-24
fix: dockerfile

tsunyoku:
# Describe your changes

Type fix PR purged `make run-prod` in favour of `make run` being the only run command, which wasn't reflected in the Dockerfile, so currently it just crashes.

## Checklist


NiceAesth (approved):

cmyui:
hmm idk about having the dockerfile depend on the makefile either, imo the makefile is for dev and not prod

cmyui:
i think makefile should probably be in .dockerignore if anything and not bundled with the application

PR #513, opened 2023-09-25
Unify manual & docker execution modes into a single docker-backed runtime

cmyui:
# Describe your changes

This PR aims to simplify the project by consolidating the two execution modes (manual & docker) into a single docker-backed runtime, and to improve & simplify the tooling, and documentation around it.

This PR additionally moves the reverse proxy out of the docker-compose setup and installs it on the host machine, adds a script to automatically configure nginx, and removes the geoip nginx module support in preference of using cloudflare headers (for which we provide documentation).

I aim to build upon this by running the docker-compose setup in CI to run integration tests against instances of the service; see #508 for a POC.

Please let me know of any concerns such that I can address them.

## Checklist


cmyui:
I am to build upon this by running the docker-compose setup in CI to run integration tests against instances of the service.

cmyui:
After this PR, running the simple is typically a series of `make build` & `make run`. The docker-compose dev environment has been fitted with auto-reload using uvicorn's `reload=True` & a volume to disk in docker-compose.

main.py
```diff
@@ -80,39 +80,37 @@ def main(argv: Sequence[str]) -> int:
...
+        host = app.settings.APP_HOST
+        port = app.settings.APP_PORT
     elif (
-        app.utils.is_valid_unix_address(app.settings.SERVER_ADDR)
-        and app.settings.SERVER_PORT is None
+        app.utils.is_valid_unix_address(app.settings.APP_HOST)
+        and app.settings.APP_PORT is None
     ):
-        uds = app.settings.SERVER_ADDR
+        uds = app.settings.APP_HOST
```
tsunyoku, line 92: if we're forcing the use of docker then we can't use unix sockets unless they manually bridge the file... shouldn't we just remove this?

requirements-dev.txt
```diff
@@ -50,6 +50,7 @@ httpx==0.25.0; python_version >= '3.8'
 idna==3.4; python_version >= '3.5'
 orjson==3.9.7; python_version >= '3.7'
 psutil==5.9.5; python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3'
+py-cpuinfo==9.0.0
```
tsunyoku, line 53: dev?

tsunyoku (commented):
i'm not too sure how i feel about "enforcing" docker usage. i don't feel strongly about either way, but i know there is a crowd of non-docker users who probably wouldn't be particularly happy

cmyui:
> i'm not too sure how i feel about "enforcing" docker usage. i don't feel strongly about either way, but i know there is a crowd of non-docker users who probably wouldn't be particularly happy

Yeah I don't foresee it being difficult to run bancho.py without docker either - I think as long as you have the virtual env setup & networking consistent, then especially w/ nginx being on the host machine, it should work both ways

requirements-dev.txt
cmyui: seems like dev is a superset of regular deps

scripts/install-nginx-config.sh
```diff
@@ -0,0 +1,16 @@
...
+set +a
+
+echo "Installing nginx configuration"
+envsubst < ext/nginx.conf.example > /etc/nginx/sites-available/bancho.conf
+ln -s /etc/nginx/sites-available/bancho.conf /etc/nginx/sites-enabled/bancho.conf
+
+echo "Restarting nginx"
+nginx -s reload
+
+echo "Nginx configuration installed with current"
```
cmyui, line 16:
```suggestion
echo "Nginx configuration installed"
```

scripts/start_server.sh
```diff
@@ -2,9 +2,9 @@
 set -euxo pipefail
 
 # Checking MySQL TCP connection
-scripts/wait-for-it.sh $DB_HOST:$DB_PORT
+scripts/wait-for-it.sh --timeout=60 $DB_HOST:$DB_PORT
```
cmyui, line 5: timeouts probably better as an env var

.env.example
```diff
@@ -1,15 +1,13 @@
...
-SERVER_PORT=80
+APP_HOST=0.0.0.0
+APP_PORT=80
 
-HOST_PORT=443
+# XXX: make sure this matches your domain
+DOMAIN=example.com
 
-# XXX: change your db credentials
+# XXX: ensure these match your db credentials
```
cmyui, line 7:
```suggestion
# XXX: make surethese match your db credentials
```

.env.example
```diff
@@ -1,15 +1,13 @@
-# for nginx reverse proxy to work through the containers, the bancho
-# server must expose itself on port 80 to be accessed on http://bancho.
-SERVER_ADDR=0.0.0.0
-SERVER_PORT=80
+APP_HOST=0.0.0.0
+APP_PORT=80
```
NiceAesth, line 2: Shouldn't this port be changed so it does not potentially conflict with nginx if exposed?

tsunyoku:
> > i'm not too sure how i feel about "enforcing" docker usage. i don't feel strongly about either way, but i know there is a crowd of non-docker users who probably wouldn't be particularly happy
> 
> 
> 
> Yeah I don't foresee it being difficult to run bancho.py without docker either - I think as long as you have the virtual env setup & networking consistent, then especially w/ nginx being on the host machine, it should work both ways

yep, i'm fine with that

would like to be less destructive with large env changes that will break users env regardless of their setup tho, we should have deprecation warnings for these things and support the old ones for a little bit until we eventually remove that warning imo, discussed on discord with niceaesth on this one

.env.example
cmyui: this port is internal to the container - just pushed up a fix to expose it on 10000 & use that in nginx on host machine

cmyui: perhaps this APP_PORT should represent the externally exposed port, since 80 should actually be const for port inside the container

cmyui: perhaps they should be the same inside & outside of the container for even more simplicity

cmyui: ok addressed in latest commit

NiceAesth: > perhaps they should be the same inside & outside of the container for even more simplicity

yeah this is what i was trying to say, mb

docker-compose.yml
```diff
@@ -12,70 +16,74 @@ services:
...
+    image: bancho:latest
+    ports:
+      - ${APP_PORT:-${SERVER_PORT}}:${APP_PORT:-${SERVER_PORT}}
     depends_on:
       mysql:
         condition: service_healthy
       redis:
         condition: service_started
     tty: true
+    init: true
```
cmyui, line 49: @osuAkatsuki/core-osu-services @osuAkatsuki/bancho-py-devs @tsunyoku this is a neat find

https://docs.docker.com/compose/compose-file/compose-file-v3/#init

cmyui: no more are the days of writing SIGTERM handlers

ext/nginx.conf.example
```diff
@@ -1,43 +1,22 @@
...
-#geoip2 /home/user/misc/GeoLite2-City.mmdb {
-	#auto_reload 5m;
-	#$geoip2_metadata_country_build metadata build_epoch;
-	#$geoip2_data_country_code default=xx source=$remote_addr country iso_code;
-	#$geoip2_data_latitude default=0.0 source=$remote_addr location latitude;
-	#$geoip2_data_longitude default=0.0 source=$remote_addr location longitude;
-#}
 upstream bancho {
-    server unix:/tmp/bancho.sock fail_timeout=0;
+    server 127.0.0.1:${APP_PORT};
```
cmyui, line 6:
```suggestion
    server 127.0.0.1:${APP_PORT};
    # NOTE: if you wish to use unix sockets,
    # server unix:/tmp/bancho.sock fail_timeout=0;
```

NiceAesth (approved):

Makefile
```diff
@@ -1,3 +1,16 @@
+build:
+	sudo chmod 777 -R .db-data/
+	docker build -t bancho:latest .
+
+run:
+	docker-compose up bancho mysql redis
```
NiceAesth, line 6: Hindsight is 20/20

Is there any reason we _want_ to specify these explicitly?
Has caused issues for users who wish to remove/add containers to their compose

PR #521, opened 2023-09-25
fix: `outputs` and `push` cannot be used together

NiceAesth:
# Describe your changes

silently failed because it does not pass the argument on false

as per: https://docs.docker.com/build/ci/github-actions/test-before-push/

## Checklist

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -9,6 +9,9 @@ on:
     paths-ignore: ["**.md", "ext/**"]
   workflow_dispatch:
 
+env:
+  PLATFORMS: linux/amd64
+
 jobs:
   build:
     runs-on: ubuntu-latest
@@ -47,7 +50,10 @@ jobs:
       - name: Push to Docker Hub
         uses: docker/build-push-action@v3
         if: ${{ github.ref == 'refs/heads/master' }}
-        run: |
-          docker image load -i /tmp/docker.tar
-          docker image tag bancho.py:latest ${{ secrets.DOCKERHUB_USERNAME }}/bancho.py:latest
-          docker push ${{ secrets.DOCKERHUB_USERNAME }}/bancho.py:latest
+        with:
+          context: .
+          platforms: $PLATFORMS
+          push: true
+          tags: ${{ secrets.DOCKERHUB_USERNAME }}/bancho.py:latest
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -31,20 +31,9 @@ jobs:
 
       - name: Build
         uses: docker/build-push-action@v3
-        if: ${{ github.ref != 'refs/heads/master' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-build.yaml
```diff
@@ -29,18 +29,25 @@ jobs:
...
         uses: actions/upload-artifact@v3
         with:
           name: docker
           path: /tmp/docker.tar
+
+      - name: Push to Docker Hub
+        uses: docker/build-push-action@v3
+        if: ${{ github.ref == 'refs/heads/master' }}
+        run: |
+          docker image load -i /tmp/docker.tar
```
NiceAesth, line 51: is this necessary? or will the image already be available from the build step @cmyui

NiceAesth: changed

NiceAesth:
also worth noting that artifacts are stored for 90 days. should that be reduced?

.github/workflows/docker-build.yaml
```diff
@@ -29,18 +32,28 @@ jobs:
...
           tags: ${{ secrets.DOCKERHUB_USERNAME }}/bancho.py:latest
           cache-from: type=gha
           cache-to: type=gha,mode=max
-          outputs: type=docker,dest=/tmp/docker.tar
 
       - name: Upload artifact
         uses: actions/upload-artifact@v3
         with:
           name: docker
           path: /tmp/docker.tar
```
cmyui, line 48: i assume the above build step ONLY builds and sends to /tmp/docker.tar? aka we still need to upload here?

.github/workflows/docker-build.yaml
```diff
@@ -29,18 +32,28 @@ jobs:
...
+      - name: Push to Docker Hub
+        uses: docker/build-push-action@v3
+        if: ${{ github.ref == 'refs/heads/master' }}
+        with:
+          context: .
+          platforms: $PLATFORMS
+          push: true
+          tags: ${{ secrets.DOCKERHUB_USERNAME }}/bancho.py:latest
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
```
cmyui, line 59: will this trigger a 2nd build? does it use the artifact?

NiceAesth: have followed https://docs.docker.com/build/ci/github-actions/test-before-push/

it will use the cached version

.github/workflows/docker-build.yaml
NiceAesth: yes, it does not push

.github/workflows/docker-build.yaml
cmyui: i see that's using `docker/build-push-action@v5` while we're using v3

should we upgrade?

NiceAesth: oh, did not notice that. yeah we should upgrade with all the recent github changes with actions and node version

NiceAesth: updated versions

cmyui (approved):

PR #524, opened 2023-09-25
fix: make `scripts/install-nginx-config.sh` executable

minisbett:
# Describe your changes

Marked `scripts/install-nginx.config.sh` as executable.

## Checklist


minisbett:
Replaced by #525 because git is weird.

PR #525, opened 2023-09-25
refactor: tools/scripts folders, getting rid of old migrations

minisbett:
# Describe your changes

Old migrations (`migrate_logs.py` and `migrate_v420`) are very outdated and can be safely removed. In case migrate_v420 is relevant, a hint towards a perma link to the folder in the current repository state has been added with a comment.

Having two separate folders for essentially the same thing (scripts and tools) seems a bit pointless and bloats the root folder. Instead, both things find their home in a `tools` folder, with a separation between `setup` and `utils`.

`setup` contains tools such as the cloudflare dns one or the local cert generator, all tools that are related to setting up the server.
`utils` currently contains just the `recalc.py` tool but is meant to contain all tools that are related to helping the server owner doing server-related things that are not necessarily development-related. They're used more frequently while the tools in `setup` are just one-time use.
The root directory of the tools folder contains the `start_server.sh` and `wait-for-it.sh` script, as they belong in neither.


tsunyoku:
also to touch on the mention of migrations - i'd like to rethink the way they're managed currently as it is inherently ass

we can theoretically move to go-migrate and convert all of the migrations although the old ones were version-bound so may require more thought

out of scope from this pr, but it just made me think of it

cc @cmyui

tsunyoku (commented):
idk why i didn't leave those as review comments

minisbett:
> i think putting wait-for-it and start_server in dev is a little disorganised as they are used in production too.
> 
> i'm happy to just have those sit in `tools` and let the other 2 folders exist alongside

Maybe rename the dev folder to smth like "setup" then? Thought about smth like that as well cuz those two are a bit out of place

cmyui:
I might revisit this/adapt some ideas from the PR, but going to close for the moment as I'm uncertain of what direction we'd like to take

PR #528, opened 2023-09-26
fix: workaround secrets not available

NiceAesth:
secrets are not available to users without write access to the repo as they are currently configured. with this change this should not be an issue anymore as if you are pushing to master you have write access anyway. we didn't need login unless we were going to build.

# Describe your changes

## Checklist

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -25,6 +25,7 @@ jobs:
 
       - name: Login to Docker Hub
         uses: docker/login-action@v3
+        if: ${{ github.ref == 'refs/heads/master' }}
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}


NiceAesth:
how exactly is it destructive? it is pushing as an artifact to the job itself

tsunyoku:
no, because it uploads to dockerhub..?

NiceAesth:
> no, because it uploads to dockerhub..?

please attempt to read the entire action i have nothing more to add lol https://github.com/osuAkatsuki/bancho.py/blob/291c0b71b499382cdda224781d62cdf601bb894a/.github/workflows/docker-build.yaml#L50C1-L50C53

NiceAesth:
step1 builds to action cache and artifact. step2 which is only ran on master pushes to hub

tsunyoku:
ok it should be fine then, i think the order of steps in this workflow is a little odd though

tsunyoku (approved):

NiceAesth:
> ok it should be fine then, i think the order of steps in this workflow is a little odd though

yeah was debating moving login right before the docker hub push, maybe another time

Issue #420, opened 2023-04-01
bug: manually ranked (frozen status) maps revert status after some time


0i8:
### Describe the bug

maps manually edited in the database to have a ranked status, or have their status changed by !map rank map/set will be set as frozen and set to a new ranked status; will revert back to unfrozen and have their status reverted to pending after some time

some time is undefined, could be 2 weeks or could be 2 months -- unsure of the cause

### To Reproduce

1. change the ranked status of a map that is currently pending/graveyarded
2. wait an undefined amount of time
3. the map's ranked status will be reverted to pending

### Expected behavior

the map should not change status while it is frozen

### bancho.py Version

4.7.2

### Python Version

3.9.X (Default)

cmyui:
This is really bad ^

cmyui:
This is intended to be handled here
https://github.com/osuAkatsuki/bancho.py/blob/a043f12e4a77f83faa180938214e02f79333872d/app/objects/beatmap.py#L482-L486

cmyui:
Ah, there's no persistence of `frozen` there, that might be it

cmyui:
Made a PR for this theory ^ but I have not yet tested it

- #475

brudermusscode:
Having the same issue here

minisbett:
Might've found the problem!
https://discord.com/channels/748687781605408908/748688897168375819/1156306492484489277

In short, `frozen` is not set to `True` in cache when using the `!map ... set` command.

TheFantasticLoki:
I'd like to mark this as not being fixed, We at kawata are experiencing the same issue when we change a maps status from our panel. We set the status and then set frozen to 1 yet maps still revert back to original status.

minisbett:
Hm that's concerning..

tsunyoku:
> I'd like to mark this as not being fixed, We at kawata are experiencing the same issue when we change a maps status from our panel. We set the status and then set frozen to 1 yet maps still revert back to original status.

beatmap cache in bancho.py is probably taking over the database status, this whole cache shit in bpy is just annoying because it makes stuff like this terrible. i might just set out to remove some of these problematic caches

PR #534, opened 2023-09-28
fix: ensure frozen state on database and cache are in sync on !map

minisbett:
# Describe your changes

When using the `!map` command to update the status of a beatmap the map is always being frozen. Currently however, the code does set frozen to true in the database, but for the local cache it only updates the status.

I noticed this when debugging #420 and this PR might offer the solution for that issue. I invalidated the cache when debugging (_cache_expired always returned true) which meant it's always being fetched from the api, which caused the status to be reverted because when checking whether it's frozen on the api check it wasn't (since it checks the cache where it was false, not the database where it's true).

## Checklist
- [X] The changes pass pre-commit checks (`make lint`)
- [X] The changes follow coding style

### DIFF
--- a/app/commands.py
+++ b/app/commands.py
@@ -655,7 +655,7 @@ async def _map(ctx: Context) -> str | None:
             for bmap in app.state.cache.beatmapset[bmap.set_id].maps:
                 bmap.status = new_status
                 bmap.frozen = True
-                
+
             # select all map ids for clearing map requests.
             map_ids = [
                 row["id"]
--- a/app/commands.py
+++ b/app/commands.py
@@ -648,13 +648,13 @@ async def _map(ctx: Context) -> str | None:
     async with app.state.services.database.connection() as db_conn:
         if ctx.args[1] == "set":
             # update all maps in the set
-            for bmap in bmap.set.maps:
-                await maps_repo.update(bmap.id, status=new_status, frozen=True)
+            for _bmap in bmap.set.maps:
+                await maps_repo.update(_bmap.id, status=new_status, frozen=True)
 
             # make sure cache and db are synced about the newest change
-            for bmap in app.state.cache.beatmapset[bmap.set_id].maps:
-                bmap.status = new_status
-                bmap.frozen = True
+            for _bmap in app.state.cache.beatmapset[bmap.set_id].maps:
+                _bmap.status = new_status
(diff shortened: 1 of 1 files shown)


tsunyoku (approved):

app/commands.py
```diff
@@ -664,17 +658,25 @@ async def _map(ctx: Context) -> str | None:
...
         else:
             # update only map
             await maps_repo.update(bmap.id, status=new_status, frozen=True)
 
             map_ids = [bmap.id]
 
+            # make sure cache and db are synced about the newest change
             if bmap.md5 in app.state.cache.beatmap:
                 app.state.cache.beatmap[bmap.md5].status = new_status
+                app.state.cache.beatmap[bmap.md5].frozen = True
```
cmyui, line 679: nit: can both if/else branches be adjusted so `map_ids` is fetched last?

e.g.
```py
update_sql()
update_cache()
map_ids = ...
```

cmyui: it's kinda a side effect of the real behaviour of the branches so it's secondary

app/commands.py
```diff
@@ -650,11 +650,14 @@ async def _map(ctx: Context) -> str | None:
 
     async with app.state.services.database.connection() as db_conn:
         if ctx.args[1] == "set":
-            # update whole set
-            await db_conn.execute(
-                "UPDATE maps SET status = :status, frozen = 1 WHERE set_id = :set_id",
-                {"status": new_status, "set_id": bmap.set_id},
-            )
+            # update all maps in the set
+            for bmap in bmap.set.maps:
```
cmyui, line 654: `bmap` is being overwritten here

app/commands.py
```diff
@@ -650,11 +650,14 @@ async def _map(ctx: Context) -> str | None:
...
+                await maps_repo.update(bmap.id, status=new_status, frozen=True)
+
+            # make sure cache and db are synced about the newest change
+            for bmap in app.state.cache.beatmapset[bmap.set_id].maps:
```
cmyui, line 658: and here

cmyui (approved):
Adjusted the variable names to not shadow anything - looks like there were actually pre-existing bugs in this regard prior to the PR 😓

Thanks for the contribution! 🎉

PR #539, opened 2023-10-06
fix: update safe name with name on !changename command

foksurek:
# Describe your changes

This PR introduces significant change to resolve specific issue. The main change is:

- **Adding 'safe_name' Field Update**: I'm introducing an update to the `safe_name` field when updating the `name` field. This resolves the issue where the 'safe_name' field was not correctly updated when using the '!changename' command.

## Checklist


NiceAesth (approved):
lgtm

NiceAesth (changes requested):
safe_name needs to be added to playerupdatefields

cmyui (approved):
Thanks for the PR! :)

PR #544, opened 2023-11-04
fix: reset slot `loaded` & `skipped` attributes upon match completion, `!mp abort` and picking a new map

Lekuruu:
# Describe your changes

I've noticed that the `loaded` and `skipped` attributes of a slot do not get reset, after finishing a map or aborting it. This could lead to some unwanted behavior.

## Checklist


minisbett (approved):

cmyui:
Made an adjustment to keep unready_players() and this separate, since they don't always overlap

cmyui:
Ty for the PR! 🎉

Lekuruu:
No problem!

PR #545, opened 2023-11-05
fix: `Beatmap.url` returning an incomplete URL

minisbett:
# Describe your changes

`Beatmap.url` returns a URL in the format `/beatmapsets/{self.set.id}/{self.id}`. However, the mode is missing here (e.g. `/beatmapsets/1/75` vs `/beatmapsets/1#osu/75`). Since we don't have an in-built way to get the string of the mode (`mode[:2]` wouldn't work since `catch` is `fruits` here), the easiest way to solve this is by simply using the `/b/{self.id}` format.

## Checklist

### DIFF
--- a/app/api/domains/cho.py
+++ b/app/api/domains/cho.py
@@ -1539,7 +1539,9 @@ class MatchChangeSettings(BasePacket):
         elif player.match.map_id == -1:
             if player.match.prev_map_id != self.match_data.map_id:
                 # new map has been chosen, send to match chat.
-                map_url = f"https://osu.{app.settings.DOMAIN}/beatmapsets/#/{self.match_data.map_id}"
+                map_url = (
+                    f"https://osu.{app.settings.DOMAIN}/b/{self.match_data.map_id}"
+                )
                 map_embed = f"[{map_url} {self.match_data.map_name}]"
                 player.match.chat.send_bot(f"Selected: {map_embed}.")
 
--- a/app/commands.py
+++ b/app/commands.py
@@ -352,10 +352,7 @@ async def recent(ctx: Context) -> str | None:
     return " | ".join(l)
 
 
-TOP_SCORE_FMTSTR = (
-    "{idx}. ({pp:.2f}pp) [https://osu.{domain}/beatmapsets/{map_set_id}/{map_id} "
-    "{artist} - {title} [{version}]]"
-)
+TOP_SCORE_FMTSTR = "{idx}. ({pp:.2f}pp) [https://osu.{domain}/b/{map_id} {artist} - {title} [{version}]]"
 
 
 @command(Privileges.UNRESTRICTED, hidden=True)
--- a/app/objects/beatmap.py
+++ b/app/objects/beatmap.py
@@ -300,7 +300,7 @@ class Beatmap:
     @property
     def url(self) -> str:
         """The osu! beatmap url for `self`."""
-        return f"https://osu.{app.settings.DOMAIN}/beatmap/{self.id}"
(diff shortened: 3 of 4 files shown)


tsunyoku:
`/beatmap` does not exist on `osu.ppy.sh` - it should be `/b`

but i know these redirects will be gone soon (they're old shit from the `old.ppy.sh`), so i don't think this is particularly a good idea

minisbett:
> `/beatmap` does not exist on `osu.ppy.sh` - it should be `/b`
> 
> but i know these redirects will be gone soon (they're old shit from `old.ppy.sh`), so i don't think this is particularly a good idea

/b is going to be removed? man
But I think short-term this is a good way to solve it in bpy, it's just a simple click on merge and when the time has come to work on this to act on the official servers' changes

tsunyoku:
> > `/beatmap` does not exist on `osu.ppy.sh` - it should be `/b`
> 
> > 
> 
> > but i know these redirects will be gone soon (they're old shit from `old.ppy.sh`), so i don't think this is particularly a good idea
> 
> 
> 
> /b is going to be removed? man
> 
> But I think short-term this is a good way to solve it in bpy, it's just a simple click on merge and when the time has come to work on this to act on the official servers' changes

or, you could add one function to map the mode to the correct string and use the current url format...

g1-1-1:
> `/beatmap` does not exist on `osu.ppy.sh` - it should be `/b`
> 
> but i know these redirects will be gone soon (they're old shit from `old.ppy.sh`), so i don't think this is particularly a good idea

i don't think the /b/ redirect won't be removed any time soon because it will just contribute to massive linkrot and i know peppy would care about something like that

minisbett:
> > `/beatmap` does not exist on `osu.ppy.sh` - it should be `/b`
> > but i know these redirects will be gone soon (they're old shit from `old.ppy.sh`), so i don't think this is particularly a good idea
> 
> i don't think the /b/ redirect won't be removed any time soon because it will just contribute to massive linkrot and i know peppy would care about something like that -- not saying your suggestions are invalid or anything but it would be weird for peppy to remove something like that with how old his game is

same + I don't see why he'd remove it, yes it might be from old.ppy.sh but it's still a good feature because you can have links without knowing a set id

tsunyoku:
this is becoming off-topic, this pr isn't a debate on if peppy will keep the old endpoint. i'd much prefer you just added a map for mode->string and used the new url format.

i say this also because the ingame direct popup on click requires the new url format, and if we move to `/b` then these won't work as the embeds we send use this property.

minisbett:
> this is becoming off-topic, this pr isn't a debate on if peppy will keep the old endpoint. i'd much prefer you just added a map for mode->string and used the new url format.
> 
> i say this also because the ingame direct popup on click requires the new url format, and if we move to `/b` then these won't work as the embeds we send use this property.

That's a good point. Would I just add a new property? Like mode.as_url or smth?

tsunyoku:
> > this is becoming off-topic, this pr isn't a debate on if peppy will keep the old endpoint. i'd much prefer you just added a map for mode->string and used the new url format.
> 
> > 
> 
> > i say this also because the ingame direct popup on click requires the new url format, and if we move to `/b` then these won't work as the embeds we send use this property.
> 
> 
> 
> That's a good point. Would I just add a new property? Like mode.as_url or smth?

something like `mode.osu_string` is good i think (use a property), i don't want to couple the mode string to a url specifically - we may use it in the future for osu api stuff

cmyui:
Updated a couple of other cases to make the PR consistent.

This is fine for now to fix the immediate issue - feel free to make a follow-up to update to use the new urls.

cmyui (approved):

PR #554, opened 2024-01-30
fix: cloudflare dns record generator

NoelleTGS:
# Describe your changes
Removed the main domain being added to the subdomains in the cloudflare DNS record generator. Importing it before this fix will cause CF to think it's a large subdomain (ex. `c.example.com.example.com`).

## Checklist


NiceAesth (approved):
you are right thanks for the contribution

PR #557, opened 2024-02-03
Per-checksum score submission locks to prevent score duplicates

cmyui:
# Describe your changes
There is an ongoing issue due to this TOCTOU bug where scores (with identical checksums) may submit more than a single time (and award duplicate pp/score/etc.) if multiple get between the start and end conditions simultaneously.

This change adds a concurrency lock around the affected area to ensure only one runs at a time.

## Checklist

### DIFF
--- a/app/api/domains/osu.py
+++ b/app/api/domains/osu.py
@@ -788,167 +788,178 @@ async def osuSubmitModularSelector(
         if not score.player.restricted:
             app.state.sessions.players.enqueue(app.packets.user_stats(score.player))
 
-    # stop here if this is a duplicate score
-    if await app.state.services.database.fetch_one(
-        "SELECT 1 FROM scores WHERE online_checksum = :checksum",
-        {"checksum": score.client_checksum},
-    ):
-        log(f"{score.player} submitted a duplicate score.", Ansi.LYELLOW)
-        return Response(b"error: no")
+    # hold a lock around (check if submitted, submission) to ensure no duplicates
+    # are submitted to the database, and potentially award duplicate score/pp/etc.
+    async with app.state.score_submission_locks[score.client_checksum]:
+        # stop here if this is a duplicate score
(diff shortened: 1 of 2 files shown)


cmyui:
cc @minisbett

app/api/domains/osu.py
```diff
@@ -788,168 +788,179 @@ async def osuSubmitModularSelector(
...
-        "SELECT 1 FROM scores WHERE online_checksum = :checksum",
-        {"checksum": score.client_checksum},
-    ):
-        log(f"{score.player} submitted a duplicate score.", Ansi.LYELLOW)
-        return Response(b"error: no")
-
-    # all data read from submission.
-    # now we can calculate things based on our data.
-    score.acc = score.calculate_accuracy()
+    # hold a lock around (check if submitted, submission) to ensure no duplicates
```
cmyui, line 791: protip to make prs like this readable

image

tsunyoku (approved):

app/api/domains/osu.py
```diff
@@ -788,168 +788,179 @@ async def osuSubmitModularSelector(
...
+    # are submitted to the database, and potentially award duplicate score/pp/etc.
+    async with app.state.score_submission_locks[score.client_checksum]:
```
cmyui, line 793: Hmm I wonder if I could put a lifespan on the lock objects -- like a timeout in redis, to avoid holding the memory until shutdown

cmyui: Doesn't seem like a release blocker imo given
```py
$ python3
Python 3.11.7 (main, Dec  8 2023, 18:56:58) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> import asyncio
>>> sys.getsizeof(asyncio.Lock())
56```

PR #559, opened 2024-02-04
Refactor: More clearly divide responsibilities of widely used `ensure_local_osu_file` function

cmyui:
# Describe your changes

## Checklist


app/api/domains/cho.py
```diff
@@ -1252,7 +1251,7 @@ async def handle(self, player: Player) -> None:
                             ]
 
                             results = app.usecases.performance.calculate_performances(
-                                osu_file_path=str(osu_file_path),
+                                osu_file_path=str(BEATMAPS_PATH / f"{bmap.id}.osu"),
```
cmyui, line 1254: i'll probably refactor this further - i think i prefer passing the beatmap id down to lower levels here

tsunyoku: meh, i don't think there's really a point to this because beatmap id is not important for calculations

cmyui: hmm yea i wonder where the level of when a beatmap id vs. beatmap path is more contextually appropriate -- i suppose it could be here

cmyui: moving it one layer down (into `calculate_performances()`) would mean it would happen at the layer where the calculator is created & method invoked - it's labelled as a `usecase`. maybe it would be more appropriate in that layer? i don't really have a strong opinion but might reduce code duplication and imports of `BEATMAPS_PATH`

NiceAesth (dismissed):
ok

NiceAesth (approved):
OK

PR #566, opened 2024-02-09
param name, type and style normalization across all repos

cmyui:
# Describe your changes

## Checklist


NiceAesth:
mypy now fails on all prs, will look into if nobody beats me to it

PR #569, opened 2024-02-09
fix: cast `__eq__` to bool

NiceAesth:
# Describe your changes

This is to appease to mypy as `__eq__` technically returns `Any`. Python casts to bool anyway so this should have 0 impact other than silencing mypy.

## Checklist


cmyui (approved):

cmyui:
tysm on the quick notice & fix for the mypy failures btw! very much appreciated

PR #571, opened 2024-02-10
chore: replace `reorder-python-imports` with `isort` to address `black` incompatibility

NiceAesth:
# Describe your changes

Can unlock black version with this

## Checklist


NiceAesth:
We can disable adding the `__future__ import annotations` import now via comments

cmyui:
What's your motivation to switch to `isort`? My choice of `reorder-python-imports` was in the vein of minimizing merge conflicts. Also tool quality i guess -- i highly regard/trust asottile haha

minisbett:
> What's your motivation to switch to `isort`? My choice of `reorder-python-imports` was in the vein of minimizing merge conflicts. Also tool quality i guess -- i highly regard/trust asottile haha

Honestly I've just googled for "python linq equivalent" and found this XD

NiceAesth:
The fact it completely breaks black and neither maintainer wants to support the other tool out of stubbornness

This also does pretty much the same import style anyway just with a couple behaviour changes

NiceAesth:
Forgot to link to this issue in the original post

https://github.com/psf/black/issues/4175

app/api/__init__.py
```diff
@@ -1,5 +1,5 @@
 # type: ignore
-from __future__ import annotations
+# isort: dont-add-imports
```
cmyui, line 2: Know if there's a way to automagically handle `__init__.py` files globally in the `.isort.cfg`?

cmyui:
> The fact it completely breaks black and neither maintainer wants to support the other tool out of stubbornness
> 
> This also does pretty much the same import style anyway just with a couple behaviour changes
> 
> * Will sort imports in stuff like `TYPE_CHECKING` blocks
> * Has support for disabling certain features per-file via comments e.g. disabling adding imports (i think reorder-python-imports also had this but on a simpler level)
> * Has configs for sort behaviour (so we can use it with black)
> 
> the other alternatives are switching to ruff or forking the tool to make it compatible and i view neither of these as necessary personally. isort seemed to be the lowest lift solution for resolving the incompatibility

Interesting okay -- damn that's unfortunate indeed, but understood. I see this PR aligns with choices made by many other familiar faces in the py ecosystem, so I'm aligned with the change.

+1 👍🏻

app/api/__init__.py
NiceAesth: seems like there is no way to do this sadly

tsunyoku (approved):

PR #574, opened 2024-02-12
fix: Wrong index for fetching global & country rank from redis on API `/get_player_info`

minisbett:
# Describe your changes

The code enumerated through the stats dictionary of the user, but used the index in that enumeration for the mode key instead of the actual mode value, which works fine... Up until autopilot, since the 7 (rx!mania) is skipped and 8 is the proper number to use. This caused the API to give 0 rank for autopilot for any user.

## Checklist


cmyui (approved):
+1 👍🏻

PR #588, opened 2024-02-13
fix: allow disabling specific calculations

NiceAesth:
# Describe your changes

Addresses issues raised in #573 
Added `tzdata` to dependencies, meant to pr separately but oh well

## Checklist


tsunyoku (approved):

tools/recalc.py
```diff
@@ -219,25 +219,26 @@ async def recalculate_mode_scores(mode: GameMode, ctx: Context) -> None:
 
 async def main(argv: Sequence[str] | None = None) -> int:
     argv = argv if argv is not None else sys.argv[1:]
-    if len(argv) == 0:
-        argv = ["--help"]
```
cmyui, line 223: does it still behave the same way after the change here?

NiceAesth: Forcing that argument on makes it impossible to run it without args (which runs on all modes all calcs)

You need to add --help now

PR #590, opened 2024-02-13
fix: scoreframe definition was invalid

kiwec:
Doesn't affect anything since it's not used anywhere, but who knows, it might be in the future.


tsunyoku (dismissed):
thanks

NiceAesth (dismissed):

cmyui (approved):

PR #597, opened 2024-02-13
Split `MapPools` and related collections/cache into repositories

cmyui:
# Describe your changes

This PR acts to serve as an example of how to split objects/collections into repositories.

In this example, the data was simply moved out of application memory and now only exists in the SQL database. I suspect the implementation for `clans` will be very similar, as clans are made up only of persistent data (clans and clan members).

The implementation of `channels` will be somewhat similar -- because similarly there are channels and channel members, however channel members are ephemeral.

The `matches` will be a bit different - in the current bancho.py model, they're entirely ephermeral data which only exists until the server reboots. They should likely be moved to redis in the long-term, but could stay in application memory and still be encapsulated in a repository. Either would be acceptable for a PR.

Splitting up `players` will be the trickiest bit, as I think we should divide the persistent and ephemeral pieces of a player up into separate repositories - the `accounts` and the `sessions`.

## Checklist

### DIFF
--- a/app/repositories/tourney_pools.py
+++ b/app/repositories/tourney_pools.py
@@ -59,7 +59,7 @@ async def create(name: str, created_by: int) -> TourneyPool:
 
 
 async def fetch_many(
-    pool_id: int | None = None,
+    id: int | None = None,
     created_by: int | None = None,
     page: int | None = 1,
     page_size: int | None = 50,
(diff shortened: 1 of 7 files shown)


app/repositories/tourney_pools.py
```diff
@@ -0,0 +1,144 @@
...
+        SELECT {READ_PARAMS}
+          FROM tourney_pools
+         WHERE id = :id
+    """
+    params: dict[str, Any] = {
+        "id": id,
+    }
+    tourney_pool = await app.state.services.database.fetch_one(query, params)
+    return (
+        None if tourney_pool is None else cast(TourneyPool, dict(tourney_pool._mapping))
```
James-Tepper, line 118: cast at beginning for consistency

cmyui:
Needs testing. Marking as draft for now.

cmyui:
Tested and working

PR #598, opened 2024-02-13
refactor: favourites repository

7mochi:
# Describe your changes

## Checklist

### DIFF
--- a/app/repositories/favourites.py
+++ b/app/repositories/favourites.py
@@ -8,13 +8,13 @@ from typing import cast
 
 import app.state.services
 
-# +--------------+------------------------+------+-----+---------+----------------+
-# | Field        | Type                   | Null | Key | Default | Extra          |
-# +--------------+------------------------+------+-----+---------+----------------+
-# | userid       | int                    | NO   | PRI | NULL    |                |
-# | setid        | int                    | NO   |     | NULL    |                |
-# | created_at   | int                    | NO   |     | 0       |                |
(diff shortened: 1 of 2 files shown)


app/repositories/favourites.py
```diff
@@ -0,0 +1,90 @@
...
+
+import app.state.services
+
+# +--------------+------------------------+------+-----+---------+----------------+
+# | Field        | Type                   | Null | Key | Default | Extra          |
+# +--------------+------------------------+------+-----+---------+----------------+
+# | userid       | int                    | NO   | PRI | NULL    |                |
+# | setid        | int                    | NO   |     | NULL    |                |
+# | created_at   | int                    | NO   |     | 0       |                |
+# +--------------+------------------------+------+-----+---------+----------------+
```
cmyui, line 17: UserID is the primary key? Doesn't this mean a user can only have a single favourite? Wtf

7mochi: i forgot setid :tf:

`primary key (userid, setid)`

cmyui: Kk, please use `desc` in the future to avoid any confusion/mis-reviews of PRs

cmyui: it slipped into the `create()` too -- the select isn't using the setid

cmyui: i'll fix -- check the diffs for some ideas perhaps

cmyui:
tested & working

PR #603, opened 2024-02-15
feat.: add setting to disable in-game registration

minisbett:
# Describe your changes

A lot of servers deactivate this because this feature is very prone to user-botting. Figured it'd make sense to allow developers to disable this via the config.

## Checklist


app/api/domains/osu.py
```diff
@@ -1667,6 +1667,18 @@ async def register_account(
     # are safe for registration.
     errors: Mapping[str, list[str]] = defaultdict(list)
 
+    # Disable in-game registration if enabled
+    if app.settings.DISABLE_INGAME_REGISTRATION:
+        errors["password"].append(
+            "The in-game registration is disabled. Please register on the website.",
+        )
+        errors = {k: ["\n".join(v)] for k, v in errors.items()}
+        errors_full = {"form_error": {"user": errors}}
```
tsunyoku, line 1676: since we're now doing this in 2 places it might be nice to extract this into a common method

app/api/domains/osu.py
```diff
@@ -1667,6 +1667,18 @@ async def register_account(
     # are safe for registration.
     errors: Mapping[str, list[str]] = defaultdict(list)
 
+    # Disable in-game registration if enabled
+    if app.settings.DISABLE_INGAME_REGISTRATION:
+        errors["password"].append(
+            "The in-game registration is disabled. Please register on the website.",
```
tsunyoku, line 1673:
```suggestion
            "In-game registration is disabled. Please register on the website.",
```

`The` doesn't really have a place here

.env.example
```diff
@@ -54,6 +54,7 @@ PP_CACHED_ACCS=90,95,98,99,100
 DISALLOWED_NAMES=mrekk,vaxei,btmc,cookiezi
 DISALLOWED_PASSWORDS=password,abc123
 DISALLOW_OLD_CLIENTS=True
+DISALLOW_INGAME_REGISTRATION=False
```
tsunyoku, line 57: i might honestly argue that the default should be `True` despite the clear change in behaviour, it being enabled has caught out a lot of servers because of botting and you should probably be aware of the risks if you're going to use this

tsunyoku (changes requested):
aligned on the idea overall, just a couple of suggestions

app/api/domains/osu.py
minisbett: I think this can be simplified, since it only returns one single error

.env.example
minisbett: yeah fair

app/api/domains/osu.py
tsunyoku: maybe we can have a constant, something like this:

```py
INGAME_REGISTRATION_DISALLOWED_ERROR = {"form_error": {"user": {"password": ["In-game registration is disabled. Please register on the website."]}}}
```

and then just return that as the content

minisbett: Check my commit

tsunyoku: it would be clearer in intentions imo, you had the same idea as me just you didn't move it into a constant

NiceAesth: it seems a bit... too weird to have this as a constant to me.

if we're doing that i'd rather store generic responses somewhere and have smth like
```py
INGAME_REGISTRATION_DISALLOWED_RESPONSE = ORJSONResponse(...)
```

tsunyoku: yeah, maybe. i just think it reads weird in-lined.

app/api/domains/osu.py
```diff
@@ -1667,6 +1677,13 @@ async def register_account(
     # are safe for registration.
     errors: Mapping[str, list[str]] = defaultdict(list)
 
+    # Disable in-game registration if enabled
+    if app.settings.DISALLOW_INGAME_REGISTRATION:
+        return ORJSONResponse(
+            content=INGAME_REGISTRATION_DISALLOWED_ERROR,
+            status_code=status.HTTP_400_BAD_REQUEST,
+        )
```
cmyui, line 1685: might as well move it above `errors` initialization.. the param validation is more debatable ig

cmyui (dismissed):

cmyui (approved):

PR #604, opened 2024-02-15
fix: ignore "0" disk-serial md5 on client hash checks

minisbett:
# Describe your changes

Some disk manufacturers do not have a disk serial set, defaulting to the number 0. This can cause lots of trouble over time as anyone with such a disk serial will not be able to verify unless manually approved by staff.

I think this check is safe to implement and does not disregard the actual purpose of this check.

I'm not sure this is the right place to implement it though.

## Checklist


tsunyoku (commented):
i definitely agree with the precedent of ignoring these problematic hardwares (i've had to account for this elsewhere) but i think we should probably not hard-code it... 

list config variable (with defaults) or a database table sounds better, just because if people find new ones then they don't have to PR every time.

i'm honestly not sure what the best way is to navigate this; it creates an issue of either people don't have these or it's not fun to add new ones (believe me there is quite a few outliers, from memory)

generally speaking, it points out why an automatic restriction policy for this is generally flawed in my eyes, but since we're a backend-only service we don't really have a place to offer a better experience (alerts or something similar) to remove the policy in-place for good manual checks

minisbett:
> it points out why an automatic restriction policy for this is generally flawed in my eyes

I mean, it doesn't automatically restrict. It just prevents verification (first login).

NiceAesth:
would prefer an allowed serials table adding on to what @tsunyoku said

perhaps we can add some default values in migrations or something

minisbett:
> list config variable (with defaults) or a database table sounds better, just because if people find new ones then they don't have to PR every time.

hmm, haven't come across any other than this one so far but fair.

tsunyoku:
> I mean, it doesn't automatically restrict. It just prevents verification (first login).

fair, i forgot. still sucks for both parties though

minisbett:
Would you rather use a database table or config? I feel like the config makes more sense, I wouldnt really put configuration values that are barely ever touched into the database.

tsunyoku:
> hmm, haven't come across any other than this one so far but fair.

it's not really a predictable thing so yeah, the issue is just that manufacturers are not forced into creating unique identifiers (if any at all) so there is many cases of these clashing identifiers that depend on what the manufacturer decided or what the OS decided. there's no real way to deterministically solve this issue, which is why i mentioned any kind of auto enacting policy is quite flawed

tsunyoku:
> Would you rather use a database table or config? I feel like the config makes more sense, I wouldnt really put configuration values that are barely ever touched into the database.

the advantage i see (probably doesn't matter in the future) is being able to modify without restarting. this wouldn't matter but bpy is stateful so it would mean yeeting sessions, multis etc. to change this. i guess it opens a can of worms about other config variables that exist, but this one in particular isn't sensitive and also not a simple toggle so i think it's justifiable.

app/api/domains/cho.py
```diff
@@ -759,12 +759,21 @@ async def handle_osu_login_request(
...
     # TODO: store adapters individually
 
+    # some disk manufacturers have "0" for their disk serial, resulting in clashing md5 hashes of the string "0".
+    # these need to be ignored since otherwise people will be prevented from verifying a lot.
+    INACTIONABLE_DISK_SIGNATURE_MD5S = ["dcfcd07e645d245babe887e5e2daa016"]
+
+    if login_data["disk_signature_md5"] not in INACTIONABLE_DISK_SIGNATURE_MD5S:
+        disk_signature_md5 = login_data["disk_signature_md5"]
+    else:
+        disk_signature_md5 = None
```
cmyui, line 769: Moved this up to a higher level -- outside of the repo -- added support for null `disk_signature_md5` to the repo layer.

app/repositories/client_hashes.py
```diff
@@ -101,8 +103,8 @@ async def create(
 async def fetch_any_hardware_matches_for_user(
     userid: int,
     running_under_wine: bool,
-    adapters: str | None = None,
-    uninstall_id: str | None = None,
+    adapters: str,
+    uninstall_id: str,
```
cmyui, line 107: bugs in master -- these are currently _required_

app/api/domains/cho.py
```diff
@@ -759,12 +759,21 @@ async def handle_osu_login_request(
 
     # TODO: store adapters individually
 
+    # some disk manufacturers have "0" for their disk serial, resulting in clashing md5 hashes of the string "0".
+    # these need to be ignored since otherwise people will be prevented from verifying a lot.
+    INACTIONABLE_DISK_SIGNATURE_MD5S = ["dcfcd07e645d245babe887e5e2daa016"]
```
cmyui, line 764: This is fine for now imo -- comments regarding having a db table or config file are indeed more flexible, so I'd be open to a follow-up PR

cmyui:
Tested and working

cmyui (approved):

PR #605, opened 2024-02-15
Add tail to make logs

minisbett:
# Describe your changes

The current `make logs` command causes it to print out the whole log since container start, which feels very awful to use. Adding `--tail 1` causes it to start all the way at the bottom, which I think is preferred.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ run-caddy:
 	caddy run --envfile .env --config ext/Caddyfile
 
 logs:
-	docker-compose logs -f bancho mysql redis
+	docker-compose logs -f bancho mysql redis --tail 1
 
 shell:
 	poetry shell
--- a/Makefile
+++ b/Makefile
@@ -11,8 +11,9 @@ run-bg:
 run-caddy:
 	caddy run --envfile .env --config ext/Caddyfile
 
+last?=1
 logs:
-	docker-compose logs -f bancho mysql redis --tail 1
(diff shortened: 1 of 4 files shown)


tsunyoku:
meh, this is pure opinion at this point -- i often like to have atleast *some* history of prior logs. i think make lets you pass parameters after the action and it gets passed to what it executes? that would be the best way to handle this

NiceAesth:
`--follow` would probably be a better default

minisbett:
> meh, this is pure opinion at this point -- i often like to have atleast _some_ history of prior logs. i think make lets you pass parameters after the action and it gets passed to what it executes? that would be the best way to handle this

make unfortunately doesn't, afaik it's awful in terms of argument parsing. You could do smth like `make logs last=100` for example I think but directly specifying it like `make logs 100` doesn't work.

If my bpy runs over a day I need to wait many minutes for the logs to reach the bottom, therefore the current command is pretty much unusable.

minisbett:
> `--follow` would probably be a better default

This is the current behavior. But as I said before, it makes the command pretty useless if u need to wait literal minutes until u see somewhat recent logs.

tsunyoku:
> You could do smth like make logs last=100 for example I think but directly specifying it like make logs 100 doesn't work.

`last=100` is fine. better than nothing. certainly better than defining my behaviour for me

cmyui (approved):
+1 👍🏻

PR #607, opened 2024-02-15
fix: add rounding to score on getscores

minisbett:
# Describe your changes

Usually PP is always rounded, but on the `/web/osu-osz2-getscores.php` endpoint it's always just casted to an int, resulting in a different PP value than what's displayed elsewhere.

## Checklist


tsunyoku (approved):
sure

cmyui (approved):

PR #611, opened 2024-02-16
Low-lift solution to move to stdlib logging

### LINKED ISSUE #266: Switch to stdlib logging module for all logging purposes
There are many reasons to do this - but generally it just provides a more standardized & fleshed out interface which is also used by many (or most) of the 3rd party libraries we'll be using - allowing our logging configuration to have more of an meaningful impact on more of a the stack.

cmyui:
# Describe your changes

This PR moves us to working ontop of the standard library's `logging` module, as opposed to the current approach of rolling our own formatting to stdout. This adds a *lot* more flexibility. I intentionally made this change as a minimal-diff to avoid sweeping changes, but they will likely follow to give us further flexibility.

## Related Issues / Projects
Resolves #266

## Checklist


NiceAesth:
oh nice this was one of the ones i had on my todo

app/api/domains/cho.py
```diff
@@ -428,7 +429,7 @@ async def handle(self, player: Player) -> None:
             t_chan.send(msg, sender=player)
 
         player.update_latest_activity_soon()
-        log(f"{player} @ {t_chan}: {msg}", Ansi.LCYAN, file=".data/logs/chat.log")
+        log(f"{player} @ {t_chan}: {msg}", Ansi.LCYAN)
```
cmyui, line 432: This is now replaced by the flexibility of logger configuration -- you can now send logs to not only standard streams, but a file, tcp/udp socket, or anything else you desire.

logging.yaml
cmyui, line 1: I probably need to include a `logging.yaml.example` in the repo, and instruct folks to `cp logging.yaml.example logging.yaml`, similarly to how users are expected to setup their `.env` on initial setup..

This will be a breaking change, so we'll need to inform our users.

tsunyoku (dismissed):

cmyui:
> going to hold off on merging until we decide how we want to handle the breaking changes in this PR. obviously a message in `#breaking-changes` as standard, but i don't want to blindly merge this solo

I've reverted the breaking changes to `.data/logs/chat.log`. I think I'm going to change `logging.yaml` to `logging.yaml.example` though -- which will require users to `cp logging.yaml.example logging.yaml` at least.. That way they can make changes to the file without having it want to auto-include in every git push.

.github/docs/wiki/Setting-up.md
```diff
@@ -17,15 +17,23 @@ sudo apt install -y docker docker-compose
...
-`.env` file. we provide an example `.env.example` file which you can use as a base.
+`.env` and `logging.yaml` files. we will provide example files for each, which
+you can use as a base and modify as you'd like.
 
 ```sh
 # create a configuration file from the sample provided
 cp .env.example .env
 
+# create a logging configuration file from the sample provided
+cp logging.yaml.example logging.yaml
```
NiceAesth, line 28: Any reason to have this as an extra setup step if we don't expect users to edit it? Could we just provide the yaml file?

pyproject.toml
```diff
@@ -73,6 +73,7 @@ respx = "0.20.2"
 tzdata = "2024.1"
 coverage = "^7.4.1"
 databases = {version = "^0.8.0", extras = ["mysql"]}
+python-json-logger = "^2.0.7"
```
NiceAesth, line 76: Is this used anywhere?

edit: I see it exists as a disabled option in the config

.github/docs/wiki/Setting-up.md
NiceAesth: I see the comment now. Sorry, on shitty wifi

pyproject.toml
cmyui: Yeahh there's a debate that it's not necessary but I think it's a frequent enough use case (structured logging) where I think it's worth having batteries-included

PR #612, opened 2024-02-16
Remove unnecessary features to decrease maintenance/complexity overhead. + Cleanup & more repo usage.

cmyui:
# Describe your changes

This PR removes a couple of features:
- Support for unix sockets and related checks
- Insurance that the running python version is (>=3.11)
- Support for deprecated `SERVER_ADDR` and `SERVER_PORT` config vars - now only `APP_x` variants exist
- Exception hooks to catch some cases and display them in more simple light
- Validations to INET addresses
- Debugging hooks under `_testing.runtime`
- CLI handling using argparse for `main.py` (was only used for `./main.py -V`)

All of these seem like unnecessary complexity requiring extra maintenance overhead with low value delivered.

It also migrates more direct queries to SQL to instead use `players_repo`, removes usage of `__all__` in modules, and more general cleanup.

## Checklist

### DIFF
--- a/app/commands.py
+++ b/app/commands.py
@@ -15,6 +15,7 @@ from collections.abc import Mapping
 from collections.abc import Sequence
 from dataclasses import dataclass
 from datetime import datetime
+from datetime import timedelta
 from functools import wraps
 from pathlib import Path
 from time import perf_counter_ns as clock_ns
@@ -1258,7 +1259,7 @@ async def server(ctx: Context) -> str | None:
 
     return "\n".join(
         (
-            f"{build_str} | uptime: {seconds_readable(uptime)}",
+            f"{build_str} | uptime: {timedelta(seconds=uptime)}",
             f"cpu: {cpu_info_str}",
             f"ram: {ram_info}",
             f"search mirror: {mirror_search_url} | download mirror: {mirror_download_url}",
--- a/app/commands.py
+++ b/app/commands.py
@@ -65,7 +65,6 @@ from app.repositories import players as players_repo
 from app.repositories import tourney_pool_maps as tourney_pool_maps_repo
 from app.repositories import tourney_pools as tourney_pools_repo
 from app.usecases.performance import ScoreParams
-from app.utils import seconds_readable
 
 if TYPE_CHECKING:
     from app.objects.channel import Channel
--- a/app/objects/player.py
+++ b/app/objects/player.py
@@ -10,7 +10,6 @@ from enum import StrEnum
 from enum import unique
 from functools import cached_property
 from typing import TYPE_CHECKING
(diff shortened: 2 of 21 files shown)


cmyui:
This is a terrible PR - sorry - moving fast today

app/commands.py
```diff
@@ -1258,7 +1259,7 @@ async def server(ctx: Context) -> str | None:
 
     return "\n".join(
         (
-            f"{build_str} | uptime: {seconds_readable(uptime)}",
+            f"{build_str} | uptime: {timedelta(seconds=uptime)}",
```
cmyui, line 1262: today i learned: str(timedelta()) does exactly what i want

app/objects/player.py
```diff
@@ -238,7 +238,6 @@ def __init__(
 
         self.id = id
         self.name = name
-        self.safe_name = self.make_safe(self.name)
```
cmyui, line 241: Moved to virtual property below

app/settings.py
```diff
@@ -2,30 +2,16 @@
...
-APP_PORT = None
-_app_port = support_deprecated_vars(
-    new_name="APP_PORT",
-    deprecated_name="SERVER_PORT",
-    until=date(2024, 1, 1),
-    allow_empty_string=True,
-)
-if _app_port:
-    APP_PORT = int(_app_port)
+APP_HOST = os.environ["APP_HOST"]
```
NiceAesth, line 13: update docker configs as well

cmyui: Should be good I think:

image

We previously deprecated SERVER_HOST and SERVER_PORT back in Sep 2023; this just fully removes support for the already-deprecated vars.

NiceAesth: https://github.com/osuAkatsuki/bancho.py/blob/4310cf0769f903c3c9d81dc9cb109986530405a1/docker-compose.yml#L89

NiceAesth: (same thing in test yml)

cmyui: Oo ty didn't notice

cmyui: Aha, in our deprecation msg we typo'd SERVER_ADDR to SERVER_HOST

NiceAesth (dismissed):
O K TO me (albeit very scary)

TheFantasticLoki:
My only gripe is the no internet check, can be very useful in certain situations. otherwise good PR

cmyui:
> My only gripe is the no internet check, can be very useful in certain situations. otherwise good PR

👍🏻 I've re-added it

NiceAesth (approved):
ok

NiceAesth (approved):

PR #619, opened 2024-02-17
Deprecate clans collection, only hold clan id references & don't cache in memory

cmyui:
# Describe your changes

## Checklist


app/utils.py
```diff
@@ -31,6 +35,10 @@ def make_safe_name(name: str) -> str:
     return name.lower().replace(" ", "_")
 
 
+def determine_highest_ranking_clan_member(members: list[Player]) -> Player:
+    return next(iter(sorted(members, key=lambda m: m["clan_priv"], reverse=True)))
```
cmyui, line 39: Dunno where this should live atm

app/api/v1/api.py
```diff
@@ -1017,14 +971,57 @@ async def api_get_pool(
...
+                    "plays": bmap.plays,
+                    "passes": bmap.passes,
+                    "mode": bmap.mode,
+                    "bpm": bmap.bpm,
+                    "cs": bmap.cs,
+                    "od": bmap.od,
+                    "ar": bmap.ar,
+                    "hp": bmap.hp,
+                    "diff": bmap.diff,
+                }
```
cmyui, line 1024: The pattern with the input var types for the `format_x_basic()` funcs was breaking -- I noticed they were only used for this function, so I decided it's not worth maintaining for v1 imo. I think we can rethink how to share formatting rules for v2.

app/commands.py
```diff
@@ -2409,13 +2402,40 @@ async def clan_info(ctx: Context) -> str | None:
...
+
+    if not clan_members:
+        # no members left, disband clan
+        await clans_repo.delete(clan["id"])
+
+        # announce clan disbanding
+        announce_chan = app.state.sessions.channels.get_by_name("#announce")
+        if announce_chan:
+            msg = f"\x01ACTION disbanded {clan_display_name}."
+            announce_chan.send(msg, sender=ctx.player, to_self=True)
```
cmyui, line 2430: In master, the disbanding announcement only happens for `!mp disband`, and not the `!mp leave` case where no members remain & the clan is implicitly disbanded. This resolves that gap.

cmyui:
Still needs manual testing

app/api/v1/api.py
```diff
@@ -1017,14 +971,57 @@ async def api_get_pool(
             status_code=status.HTTP_404_NOT_FOUND,
         )
 
+    pool_creator_clan = await clans_repo.fetch_one(id=pool_creator.clan_id)
+    pool_creator_clan_members: list[players_repo.Player] = []
+    if pool_creator_clan:
```
cmyui, line 976:
```suggestion
    if pool_creator_clan is not None:
```

app/api/v1/api.py
```diff
@@ -1017,14 +971,57 @@ async def api_get_pool(
...
+                "name": pool_creator.name,
+                "country": pool_creator.geoloc["country"]["acronym"],
+                "clan": (
+                    {
+                        "id": pool_creator_clan["id"],
+                        "name": pool_creator_clan["name"],
+                        "tag": pool_creator_clan["tag"],
+                        "members": len(pool_creator_clan_members),
+                    }
+                    if pool_creator_clan
```
cmyui, line 997:
```suggestion
                    if pool_creator_clan is not None
```

cmyui (commented):
Tested and working

PR #624, opened 2024-02-19
Fix migrations not running if never run before

cmyui:
# Describe your changes

## Checklist


cmyui:
Yikes pretty bad one

PR #627, opened 2024-02-19
fix: catch pp wrong due to both acc and judgements specified

minisbett:
# Describe your changes
PP breaks on catch if both the accuracy and the 50s are specified. In VC we argued the best way to solve this is raising an exception in the usecase, and dropping specifying acc where possible, since it's a useless info.

This PR fixes catch PP, meaning bancho.py hosters will need to re-calculate all PP. Before doing this, I'd suggest looking into #573, as it improves the functionality of the recalc script and might be significant with this change.
## Checklist


app/usecases/performance.py
```diff
@@ -65,16 +65,8 @@ def calculate_performances(
...
-        # if (score.acc is None) == (
-        #     score.n300 is None
-        #     and score.n100 is None
-        #     and score.n50 is None
-        #     and score.ngeki is None
-        #     and score.nkatu is None
-        #     and score.nmiss is None
-        # ):
-        #     raise ValueError("Either acc OR 300/100/50/geki/katu/miss must be present")
+        if score.acc and (score.ngeki or score.nkatu or score.n50 or score.n100):
```
cmyui, line 68: n300?

tsunyoku:
bug fixes like this are really good candidates for adding unit tests to cover imo

tools/recalc.py
```diff
@@ -66,7 +66,6 @@ async def recalculate_score(
     calculator = Calculator(
```
cmyui, line 66: can we have a similar check/err raise here?

cmyui: or even better, could we make the recalc.py script use the usecase function?

tsunyoku: this should probably be using the usecase, no?

minisbett: > an we have a similar check/err raise here?
we don't need one, we simply just don't specify the acc here, why would we check for anything

app/usecases/performance.py
minisbett: When does it make sense to specify n300? With score.acc it's estimated, and when u specify 100s etc. n300 is just the leftover.

cmyui: Yeah I agree - I think we should _block_ usage of `n300` and `acc` together, the same as with geki/katu/n50/n100.

More specifically, outside of misses, any fields which affect `acc`:

image

tools/recalc.py
cmyui: Oh I see, this should be using the usecase yeah

app/usecases/performance.py
```diff
@@ -60,21 +61,24 @@ def calculate_performances(
...
-        #     and score.n50 is None
-        #     and score.ngeki is None
-        #     and score.nkatu is None
-        #     and score.nmiss is None
-        # ):
-        #     raise ValueError("Either acc OR 300/100/50/geki/katu/miss must be present")
+        if score.acc and (
+            score.n300 or score.n100 or score.n50 or score.ngeki or score.nkatu
+        ):
+            raise ValueError("Must not specify both accuracy and geki/katu/50s/100s.")
```
cmyui, line 81:
```suggestion
            raise ValueError("Must not specify accuracy AND 300/100/50/geki/katu. Only one or the other.")
```

tools/recalc.py
cmyui: Follow-up PR is fine

cmyui (dismissed):

app/usecases/performance.py
minisbett: I think we can just remove the n300 param completely? I dont think it ever has a usecase

cmyui (approved):

PR #628, opened 2024-02-19
base telemetry idea

cmyui:
# Describe your changes

## Checklist

### DIFF
--- a/app/telemetry.py
+++ b/app/telemetry.py
@@ -1,0 +1,134 @@
+from __future__ import annotations
+
+import dataclasses
+import hashlib
+import json
+import platform
+import time
+from dataclasses import dataclass
+from typing import Any
+
+import psutil
+from httpx import AsyncClient
+from tenacity import retry
+from tenacity.stop import stop_after_attempt
+
+from app import settings
+
+MIN_SLOW_QUERY_SECONDS = 5.0
+
+HTTP_CLIENT = AsyncClient()
+
+
+@dataclass
+class SoftwareInfo:
+    version: str
+    domain: str
+
+
+@dataclass
+class SystemInfo:
+    system: str
+    node: str
+    release: str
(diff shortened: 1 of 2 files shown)


app/telemetry.py
```diff
@@ -0,0 +1,134 @@
...
+        async def wrapper(*args, **kwargs):
+            start_time = time.perf_counter()
+            response = await func(*args, **kwargs)
+            end_time = time.perf_counter()
+
+            seconds_elapsed = end_time - start_time
+
+            if seconds_elapsed >= MIN_SLOW_QUERY_SECONDS:
+                event_data = {
+                    "query": args[0],
```
tsunyoku, line 118:
```suggestion
                    "query": kwargs.get("query", args and args[0]),
```

tsunyoku (commented):
i know this is drafted but i'm fairly assuming that's mostly likely because the telemetry server needs to be made, but i'm aligned on this

app/telemetry.py
tsunyoku: same shit we noticed during sqlalchemy testing, if i really care then we could do this properly:

```py
query = kwargs.get("query", None)
if query is None:
    assert args
    query = args[0]
```

or something like that, but i don't think it matters very much lol

cmyui:
> i know this is drafted but i'm fairly assuming that's mostly likely because the telemetry server needs to be made, but i'm aligned on this

Yep -- I'd also like to meditate a bit more on the past, in terms of what cases have been most in-need of my attention, and what data would have been most useful in their investigation.. Also want to consider things from a privacy POV -- haven't had too much time to think about all of this yet.

cmyui:
I think the data is the most important topic, and we could make the telemetry server pretty effortlessly once we're aligned on what the exact purposes of this system are to be.

PR #632, opened 2024-02-19
[v5.1.0] Refactor the rest of the repositories to use sqlalchemy core 1.4

### LINKED ISSUE #394: bug: COALESCE function causes slow query
### Describe the bug

repositories that uses mysql coalesce function will NOT hit index, which causes slow query and performance dropped.
here is the comparison, query with coalesce is obviously cost disastrous more time
(our server has make map_md5 as index in scores table)
slow
fast

### To Reproduce

1. players played normally in the server
2. player reported that server is slow and laggy
3. check server, mysqld process has full cpu usage
4. check mysql slow query log and it shows many slow querys

### Expected behavior

find a way to make coalesce hits index or don't use coalesce anymore

### bancho.py Version

4.7.2

### Python Version

3.9.X (Default)

### Relevant log output

```shell
An example extry of mysql slow query log
# User@Host: gulag[gulag] @ localhost [::1]  Id:    47
# Query_time: 5.227328  Lock_time: 0.000079 Rows_sent: 0  Rows_examined: 1176231
SET timestamp=1674828621;
SELECT id, map_md5, score, pp, acc, max_combo, mods, n300, n100, n50, nmiss, ngeki, nkatu,
grade, status, mode, play_time, time_elapsed, client_flags, userid, perfect, online_checksum

          FROM scores
         WHERE map_md5 = COALESCE('7bdcf5615400d8499b21c24c26fdb360', map_md5)
           AND mods = COALESCE(NULL, mods)
           AND status = COALESCE(2, status)
           AND mode = COALESCE(0, mode)
           AND userid = COALESCE(5347, userid);
```

### Additional context

_No response_

cmyui:
Co-authored-by: James Wilson <tsunyoku@users.noreply.github.com>

# Describe your changes

## Related Issues / Projects
Resolves #394

## Checklist

### DIFF
--- a/app/adapters/database.py
+++ b/app/adapters/database.py
@@ -9,6 +9,8 @@ from sqlalchemy.dialects.mysql.mysqldb import MySQLDialect_mysqldb
 from sqlalchemy.sql.compiler import Compiled
 from sqlalchemy.sql.expression import ClauseElement
 
+from app import settings
+
 
 class MySQLDialect(MySQLDialect_mysqldb):
     default_paramstyle = "named"
@@ -36,7 +38,8 @@ class Database:
             dialect=DIALECT,
             compile_kwargs={"render_postcompile": True},
         )
-        print(str(compiled), compiled.params)
+        if settings.DEBUG:
+            print(str(compiled), compiled.params)
         return str(compiled), compiled.params
 
     async def fetch_one(
--- a/app/adapters/database.py
+++ b/app/adapters/database.py
@@ -32,7 +32,10 @@ class Database:
         await self._database.disconnect()
 
     def _compile(self, clause_element: ClauseElement) -> tuple[str, MySQLParams]:
-        compiled: Compiled = clause_element.compile(dialect=DIALECT)
+        compiled: Compiled = clause_element.compile(
+            dialect=DIALECT,
+            compile_kwargs={"render_postcompile": True},
+        )
         print(str(compiled), compiled.params)
         return str(compiled), compiled.params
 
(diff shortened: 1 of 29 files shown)


tsunyoku:
@cmyui should we do alembic stuff in this PR or another one? if the latter, i want to merge this into master

tsunyoku:
also, i was thinking we could create a database adapter that wraps around `databases.Database` to take ` ClauseElement` and we do the query compilation stuff there so that we're not doing `compiled = stmt.compile(dialect=DIALECT)`, `str(compiled)` and `compiled.params` *everywhere*

cmyui:
> also, i was thinking we could create a database adapter that wraps around databases.Database to take a ClauseElement and we do the query compilation stuff there so that we're not doing compiled = stmt.compile(dialect=DIALECT), str(compiled) and compiled.params everywhere

Yup, that's a good idea

tsunyoku:
sorry for the loaded commit - was unavoidable. latest commit adds a wrapper around databases as discussed above, and additionally ensures all of our queries are compatible with sqlalchemy 2.0 (all of the selects needed a slight syntax change)

squashed quite a few bugs in the process; there were a couple random ones but the majority were the sqlalchemy statements being incorrectly passed to the database

cmyui:
Alright did my most thorough testing in a while on this PR. It still likely has some diffs in terms of the existing db schema vs. what this would generate, but I don't think anything major -- just things like incorrect `server_default`. I'll tackle those in an upcoming PR, but from a runtime perspective this looks to be working as expected for all major flows (score fetch, submit, multiplayer, spectator, etc.) and commands.

cmyui:
Ty for your support on this PR @tsunyoku!! Much appreciated 🎉🎉🎉

PR #635, opened 2024-02-25
fix: use all scores (not top 100) in recalc.py

### LINKED ISSUE #299: [BUG] Overall weighted acc calculation returning wrong values

kingdom5500:
# Describe your changes

make `tools/recalc.py` consistent with the new pp/acc calc behaviour

## Related Issues / Projects

closes #299 again

## Checklist

### DIFF
--- a/tools/recalc.py
+++ b/tools/recalc.py
@@ -134,15 +134,13 @@ async def recalculate_user(
     if not total_scores:
         return
 
-    top_100_pp = best_scores[:100]
-
     # calculate new total weighted accuracy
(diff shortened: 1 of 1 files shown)


NiceAesth (approved):

minisbett:
nice thx

PR #640, opened 2024-02-26
feat/refactor: clean-up !with parameter parsing & add more param support

minisbett:
# Describe your changes

1. Simplifies the parsing
2. Adds support for specifying n100, n50, ngeki and nkatu
3. Adds tests for the parsing

## Checklist


app/commands.py
```diff
@@ -480,41 +466,16 @@ async def _with(ctx: Context) -> str | None:
...
+        return str(score_args)
 
     result = app.usecases.performance.calculate_performances(
         osu_file_path=str(BEATMAPS_PATH / f"{bmap.id}.osu"),
         scores=[score_args],  # calculate one score
     )
 
-    return "{msg}: {pp:.2f}pp ({stars:.2f}*)".format(
-        msg=" ".join(msg_fields),
+    return "{pp:.2f}pp ({stars:.2f}*)".format(
```
cmyui, line 478: I think we should keep the `msg` here is nice in the sense that from in-game, it's easy to tell whether or not the server correctly parsed the params you send

It'll be quite nice now that the parsing is split into a function -- we can have a `format__with__args(score_args) -> str` and test that similarly

PR #641, opened 2024-02-26
feat: support mcosu clients

kiwec:
# Describe your changes

Added some configuration options so server owners can easily tell McOsu clients how to behave.  
Defaults match how the clients behave when a server doesn't signal anything: you can connect, but not submit scores.

## Related Issues / Projects

https://mcosu.kiwec.net

I'm releasing the score submission build tomorrow - code is already available on the v34.02 branch.

## Checklist

### DIFF
--- a/app/api/domains/cho.py
+++ b/app/api/domains/cho.py
@@ -195,7 +195,9 @@ async def bancho_handler(
             headers={"cho-token": "no"},
             content=(
                 app.packets.login_reply(LoginFailureReason.AUTHENTICATION_FAILED)
-                + app.packets.notification("McOsu clients are not allowed on this server.")
+                + app.packets.notification(
+                    "McOsu clients are not allowed on this server.",
+                )
             ),
         )
 


tsunyoku:
before going into the debate of if we want to support this or not.

let's say we do. is the same header present on score submission? also, how can we assure that a user isn't going to just modify the source code to not pass these headers? i have a few concerns about allowing these clients to connect at all -- and regardless of if we agree to allowing them to be supported in the code, how are we going to enforce literally anything? it's an open source client without any real guarantees (lack of anticheat etc.) so we are kind of stuck in a rock between a hard place here

tsunyoku:
yeah i understand, i'm just afraid that we're opening a *bit* of a can of worms here...

app/api/domains/cho.py
```diff
@@ -199,7 +211,10 @@ async def bancho_handler(
 
         return Response(
             content=login_data["response_body"],
-            headers={"cho-token": login_data["osu_token"]},
+            headers={
+                "cho-token": login_data["osu_token"],
+                "x-mcosu-features": f"submit={1 if app.settings.ALLOW_MCOSU_SCORES else 0}",
```
cmyui, line 216: Would more features be added in the future, and what's the serialization format for these if so?

Could we extract that into a function? (e.g. `def write_mcosu_features_header(allow_submission: bool) -> str: ...`)

minisbett:
> You can't. The point of this PR is to let the clients who _do_ pass the headers respect the admin's wishes.
Do you mean this as in that McOsu by default is fully capable of connecting to private servers, and this layer of control would provide the best it could in terms of controlling McOsu's behavior?

minisbett:
> It will give players the freedom to play the way they want, not the way peppy wants.
saying peppy decides the direction lazer goes is kinda crazy ngl

tsunyoku:
> osu!lazer is open source, osu!stable had its source code leaked also
they also enforce a client-side anticheat. this is not a fair comparison

NiceAesth:
As far as I am concerned, this is very low priority for review but I would have appreciated changes like this which introduce new behaviour to be discussed beforehand.

Keeping that in mind, I do believe that this does not fall under our responsibility and I very much disagree with the approach of introducing new client-specific checks.

Calemy:
> "EZPPFarm was the only one to approve of score submission, but the others have approved the rest of multiplayer features (hence the default settings)."

with all due respect to you and the owner of EZPPFarm, me and probably others are geniuenly not really surprisedthat they agreed. They're i think the most outlying server out of all private servers. They already do custom clients and are overall not really a good orientation to what other servers think.

WIth this being said, while i really understand and appreciate the thought of making more clients accessable to the game and servers, i do not think it's a good fit for here because of the concerns raised during this discussion.

minisbett:
Me and a maintainer briefly discussed this, and figured it might make more sense to implement this kind of support via an existing endpoint, rather than dedicating the functionality in banchopy. It is supposed to be a pretty general osu! server implementation, without any special shenanigans for some custom clients.

IMO, consider using something like the [lastfm.php endpoint](https://github.com/osuAkatsuki/bancho.py/blob/ae42bf66110484b3cb98b0f99278f297eda160cf/app/api/domains/osu.py#L266), and send a flag for someone using McOsu with score-submission or similar in there so that it can be seemlessly blocked there, without adding a lot of fanfare.
e.g.
```py
if flags & (LastFMFlags.MCOSU_SCORE_SUB | LastFMFlags.MCOSU_MULTI): # or maybe even a general LastFMFlags.MCOSU
    player.logout()
```

To me personally that'd seem like an acceptable approach, and makes private servers capable of blocking McOsu easily.
Maybe the maintainers can leave their opinion about this here.

PR #644, opened 2024-02-28
fix: docker failing to install cryptography

7ez:
# Describe your changes
since cryptography is now in rust, it will fail to install without a rust compiler installed.

## Checklist


NiceAesth:
To look into: https://cryptography.io/en/latest/installation/#static-wheels

PR #648, opened 2024-03-13
add level and level_progress to stats api routes

HorizonCode:
# Describe your changes
- Adds level and level_progress to stats routes

## Checklist


minisbett:
I'm not sure whether this is a concern for bpy or whether a frontend should implement this itself

tsunyoku:
level is a statistic so i don't have an issue with this being provided via API

7ez:
tbh would be cooler if level and level progress would be combined

tsunyoku:
i'm not convinced we need all 3. level being a float with the decimals being the progress should be enough

i don't particularly care which we pick, but either have level and progress separated or have them combined into one float. but not both

HorizonCode:
then i would say keep level and level progress, bc for frotend dev its easier to use the values

app/constants/level.py
```diff
@@ -0,0 +1,141 @@
+from __future__ import annotations
+
+import math
+
+LEVEL_GRAPH = [
```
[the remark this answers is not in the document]
tsunyoku: this is how osu! actually implements it, functions have been written but only approximate - should probably stay this way

cmyui, line 5:
```suggestion
LEVEL_GRAPH: list[int] = [
```

app/constants/level.py
```diff
@@ -0,0 +1,141 @@
...
+    26931190829,
+]
+
+
+def get_required_score_for_level(level: int) -> int:
+    if level <= 0:
+        return 0
+    if level <= 100:
+        return LEVEL_GRAPH[level - 1]
+    return LEVEL_GRAPH[99] + 100000000000 * int(level - 100)
```
cmyui, line 114:
```suggestion
    return LEVEL_GRAPH[99] + 100_000_000_000 * int(level - 100)
```

app/constants/level.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+def get_level(score: int) -> int:
+    if score <= 0:
+        return 1
+
+    if score >= LEVEL_GRAPH[99]:
+        return 100 + int((score - LEVEL_GRAPH[99]) / 100000000000)
```
cmyui, line 122:
```suggestion
        return 100 + int((score - LEVEL_GRAPH[99]) / 100_000_000_000)
```

app/constants/level.py
```diff
@@ -0,0 +1,141 @@
...
+
+    for idx, v in enumerate(LEVEL_GRAPH, start=0):
+        if v > score:
+            return idx
+
+    return 1
+
+
+def get_level_precise(score: int) -> float:
+    baseLevel = get_level(score)
```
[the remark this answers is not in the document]
cmyui: we should lint/autofix this if possible, maybe there's a pre-commit hook for it

cmyui:
left some feedback and can review more deeply later, but provided the comments are addressed i think this is good

cmyui:
would it be possible to add this functionality to the v2 api as well? i'd like to get it into a better place so web implementations begin to switch over. eventually would be good to deprecate v1 once v2 has surpassed it's capabilities, get us into a much better state for maintainability of b.py

PR #668, opened 2024-06-26
Replace use of `docker-compose` with `docker compose`.

RealistikDash:
# Describe your changes
The use of `docker-compose` has been officially deprecated in favour of `docker compose` (part of the official Docker package). On newer distros, `docker-compose` no longer works.

## Checklist

### DIFF
--- a/.github/docs/wiki/Setting-up.md
+++ b/.github/docs/wiki/Setting-up.md
@@ -10,8 +10,7 @@ git clone https://github.com/osuAkatsuki/bancho.py
 cd bancho.py
 
 # install docker for building the application image
-# install docker-compose for running the application & dependencies
-sudo apt install -y docker docker-compose
+sudo apt install -y docker
 ```
 
 ## configuring bancho.py
--- a/.github/docs/wiki/locale/de-DE/Setting-up-de-DE.md
+++ b/.github/docs/wiki/locale/de-DE/Setting-up-de-DE.md
@@ -10,8 +10,7 @@ git clone https://github.com/osuAkatsuki/bancho.py
 cd bancho.py
 
 # install docker for building the application image
-# install docker-compose for running the application & dependencies
-sudo apt install -y docker docker-compose
+sudo apt install -y docker
 ```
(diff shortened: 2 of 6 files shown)


tsunyoku:
we probably want to change the setup instructions to not use `apt` for docker as a pre-requisite. from experience, some distros/vendors/etc. do not provide a new enough docker version for us to use

tsunyoku:
another thing, WSL users should be using docker desktop and not installing docker inside of their WSL - something else we should make clear

unrelated to this though

NiceAesth:
> we probably want to change the setup instructions to not use `apt` for docker as a pre-requisite. from experience, some distros/vendors/etc. do not provide a new enough docker version for us to use

could probably just link them to the docker documentation and call it a day, it has instructions on how to use the docker install script

marcandreher:
I have the same issue `docker-compose` doesn't work anymore, for my instance I switched that. On Ubuntu 24.04 docker-compose is not working anymore

cmyui:
> could probably just link them to the docker documentation and call it a day

+1 to keep the instructions up to date

cmyui:
If this breaks someone's instance due to an old install, what's the fix for them?

NiceAesth:
> If this breaks someone's instance due to an old install, what's the fix for them? (is there a simple command i can copy/pasta?)

hmm, best option would be to get them to update their docker install but

https://docs.docker.com/compose/install/linux/ 

it is also possible to 'retrofit' the compose plugin into an existing docker install 😅 (keep in mind that most recent versions come with it pre-installed)

cmyui:
> hmm, best option would be to get them to update their docker install but
>
> https://docs.docker.com/compose/install/linux/
>
> it is also possible to 'retrofit' the compose plugin into an existing docker install 😅 (keep in mind that most recent versions come with it pre-installed)

Sweet, thanks for that -- works for me

cmyui (approved):

cmyui:
Thanks for the PR!

Issue #678, opened 2024-09-04
bug: stats updating to wrong mode due to unbelievable gm_stats in score submission


TrueRou:
### Describe the bug

Player stats updating to wrong mode randomly, often occurred when playing varities of rulesets and switching among them quickly in osu client.

This may relate to **score.player.gm_stats** returns stats of a wrong ruleset.

gm_stats comes from `self.stats[self.status.mode]` of the player, but `self.status.mode` is **unbelievable** because `self.status` gethers its data from osu!client packets. In some circumstances, `self.status.mode` can point to a ruleset different from the submitted score (`score.mode.value`) in `stats_repo.partial_update`, this can lead to a stats copy.

In legal usecases, network latency and quickly switching among rulesets can lead to this problem. AND what's more, illegal packets can be constructed to copy stats on purpose.

### To Reproduce

- Randomly happened in legal gameplays (may due to network latency).
- Can be illegally constructed and used to copy stats to another rulesets.

### Expected behavior

- Don't fully trust any packets from clients.
- Score submission in database should not depend on player session context (fields like player.status).

### bancho.py Version

5.2.2

### Python Version

3.11.X (Default)

GenaroSalomone:
try this
https://gofile.io/d/yMHabA
select the desired compiler in the installer

PR #679, opened 2024-09-04
fix: stats updating to wrong mode due to unbelievable gm_stats in score submission

TrueRou:
# Describe your changes

fix: stats updating to wrong mode due to unbelievable gm_stats in score submission

## Related Issues / Projects

Issue: #678 

## Checklist


tsunyoku (approved):
not sure why it was ever different

PR #687, opened 2024-12-26
feat: add !clan join command

nowoemi:
# Describe your changes
add a !clan join command.

## Checklist

### DIFF
--- a/app/commands.py
+++ b/app/commands.py
@@ -2440,7 +2440,7 @@ async def clan_leave(ctx: Context) -> str | None:
 
 # TODO: !clan inv, !clan leave
 
-@clan_commands.add(Privileges.UNRESTRICTED, aliases=["J"])
+@clan_commands.add(Privileges.UNRESTRICTED, aliases=["j"])
 async def clan_join(ctx: Context) -> str | None:
     """Lets a user join a Clan."""
     if not ctx.args:
--- a/app/commands.py
+++ b/app/commands.py
@@ -2438,8 +2438,36 @@ async def clan_leave(ctx: Context) -> str | None:
     return f"You have successfully left {clan_display_name}."
 
 
-# TODO: !clan inv, !clan join, !clan leave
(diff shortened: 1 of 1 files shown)


NiceAesth:
i'm not sure if allowing anyone to join any clan is a good idea. the way i thought this would work is that you would invite someone if you have the required permissions then the join command would work as the "accept" invite command

app/commands.py
```diff
@@ -2438,8 +2438,36 @@ async def clan_leave(ctx: Context) -> str | None:
     return f"You have successfully left {clan_display_name}."
 
 
-# TODO: !clan inv, !clan join, !clan leave
+# TODO: !clan inv, !clan leave
 
+@clan_commands.add(Privileges.UNRESTRICTED, aliases=["J"])
```
7ez, line 2443: I’d change the “J” to lowercase tbh

7ez:
> i'm not sure if allowing anyone to join any clan is a good idea. the way i thought this would work is that you would invite someone if you have the required permissions then the join command would work as the "accept" invite command

maybe a `!clan invite` (along with `!clan invite regen(erate)`) which would allow the owner to generate (potentially stored in the clan obj) a random string and anyone who’d want to join would just do `!clan join abc123`?

app/commands.py
noemiowo: changed to lowercase

noemiowo:
> > i'm not sure if allowing anyone to join any clan is a good idea. the way i thought this would work is that you would invite someone if you have the required permissions then the join command would work as the "accept" invite command
> 
> maybe a `!clan invite` (along with `!clan invite regen(erate)`) which would allow the owner to generate (potentially stored in the clan obj) a random string and anyone who’d want to join would just do `!clan join abc123`?

this could also be an idea, but that would require some more database modification to store clan invites and/or possibly ingame notifications of getting an invite, where the database part would be too much knowledge needed for me to add it, unless you'd have an idea on how to add it?

PR #691, opened 2025-02-16
feat: lowercase clantags by default

nowoemi:
a little QoL change for people that'd like to use it by default

# Describe your changes
Added a small change in the `!clan create` command to have lowercase clantags by default

## Checklist

### DIFF
--- a/app/commands.py
+++ b/app/commands.py
@@ -2311,7 +2311,7 @@ async def clan_create(ctx: Context) -> str | None:
     if await clans_repo.fetch_one(tag=tag):
         return "That tag has already been claimed by another clan."
 
-    if await clans_repo.fetch_one(tag=tag).upper():
+    if await clans_repo.fetch_one(tag=tag) == tag.upper():
         return "That tag has already been claimed by another clan."
 
     # add clan to sql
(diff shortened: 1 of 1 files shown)


nowoemi:
lets ignore my stupidity before

app/commands.py
```diff
@@ -2311,6 +2311,9 @@ async def clan_create(ctx: Context) -> str | None:
     if await clans_repo.fetch_one(tag=tag):
         return "That tag has already been claimed by another clan."
 
+    if await clans_repo.fetch_one(tag=tag) == tag.upper():
```
7ez, line 2314: you’re comparing a dict to a string here

7ez: also this should be clan_tag.lower() == tag.lower() probably

7ez: wait you just cloned the check

7ez (changes requested):

app/commands.py
7ez: I need sleep

minisbett:
I think upper-cased clan tags are preferred as acronyms / tags are commonly in upper-case across the world.

app/commands.py
minisbett, line 2314: This check adds.. nothing? The same check already exists above and should be used. If that check for whatever reason does not work, it should be fixed.

minisbett (changes requested):

PR #693, opened 2025-03-06
chore: update type of pp as a result of b20250306.2

mayacopeland:
# Describe your changes
Change the type of the performance in functions that create the "USER_STATS" packet

## Related Issues / Projects
N/a 
## Checklist
- [] I've manually tested my code


mayacopeland:
Saw the recent client changelog - https://osu.ppy.sh/home/changelog/stable40/20250306.2

Looks like its uint16 now.

NiceAesth (approved):

NiceAesth:
older clients should be fine with this change as well (at least up until the cap gets hit) so i don't see the need for versioning this packet

cmyui (approved):

PR #694, opened 2025-03-09
feat: remove 16-bit PP limit workaround follwing stable client update

minisbett:
# Describe your changes

## Checklist


NiceAesth:
there is no reason to remove it, it was increased to 65k in the previous PR. going past that will still break

cmyui:
> there is no reason to remove it, it was increased to 65k in the previous PR. going past that will still break

See https://github.com/osuAkatsuki/bancho.py/pull/693

PR #695, opened 2025-04-10
fix: docker composes providing wrong env arg for mysql port

7ez:
# Describe your changes

## Checklist

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -10,7 +10,7 @@ services:
       MYSQL_PASSWORD: ${DB_PASS}
       MYSQL_DATABASE: ${DB_NAME}
       MYSQL_HOST: ${DB_HOST}
-      MYSQL_PORT: ${DB_PORT}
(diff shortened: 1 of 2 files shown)


cmyui:
what's the failure case? how'd you discover it? from my testing it worked so i wonder what the difference was?

7ez:
> what's the failure case? how'd you discover it? from my testing it worked so i wonder what the difference was in our envs?

wait for it was timing out for atlas after changing port

Lekuruu:
Yeah, I experienced the same thing

7ez:
> what's the failure case? how'd you discover it? from my testing it worked so i wonder what the difference was in our envs?

I reckon you never changed the port which is what causes it to never run

cmyui:
> > what's the failure case? how'd you discover it? from my testing it worked so i wonder what the difference was in our envs?
> 
> I reckon you never changed the port which is what causes it to never run

Ahh yup that's probably it

cmyui (approved):
+1 👍🏻

Issue #631, opened 2024-02-19
bug: Fix flakiness of "Test Application Code" CI job with regard to establishing mysql connection


cmyui:
We get this output:
```
Run docker-compose -f docker-compose.test.yml exec -T bancho-test /srv/root/scripts/run-tests.sh
Initializing database...
ERROR 2002 (HY000): Can't connect to server on 'mysql-test' (115)
Error: Process completed with exit code 1.
```

Here is a sample job: https://github.com/osuAkatsuki/bancho.py/actions/runs/7965367862/job/21744744492

PR #697, opened 2025-04-30
Disable test CI job as it's extremely flaky

cmyui:
https://github.com/osuAkatsuki/bancho.py/issues/631


cmyui:
Removed from branch protection rules

PR #703, opened 2025-04-30
fix: reenable test CI job and fix mysql wrong host

7mochi:
# Describe your changes

## Checklist


NiceAesth (approved):
Change is correct, test fails should be fixed in another PR

PR #706, opened 2025-05-02
refactor: split domains

7mochi:
# Describe your changes

## Checklist


app/api/bancho/users.py
cmyui, line 1: where were the contents of cho.py migrated to? that's the part i'd expect to have under bancho/

7mochi: what I migrated was osu.py, cho.py is still pending, in bancho/users.py currently there is only the register_account method that was in osu.py, should it be in web/instead of bancho/?

cmyui:
This is still a good idea btw, keep this PR in draft, I will revisit at some point. I'll probably close this one but credit you  / merge these commits with the work for my follow-up changes which use this work.

PR #715, opened 2025-07-11
fix: tourney stream handling in get_allowed_client_versions

TrueRou:
# Describe your changes

tourney client can't connect to bancho due to old clients, as https://osu.ppy.sh/api/v2/changelog?stream=tourney has an empty builds list.

in fact, tourney client use the same channel as the its base osu stream(stable).

## Checklist

### DIFF
--- a/app/api/domains/cho.py
+++ b/app/api/domains/cho.py
@@ -728,17 +728,16 @@ async def handle_osu_login_request(
             ),
         }
 
-    if osu_version.stream is OsuStream.TOURNEY and not (
-        user_info["priv"] & Privileges.DONATOR
-        and user_info["priv"] & Privileges.UNRESTRICTED
-    ):
-        # trying to use tourney client with insufficient privileges.
-        return {
-            "osu_token": "no",
-            "response_body": app.packets.login_reply(
-                LoginFailureReason.AUTHENTICATION_FAILED,
-            ),
-        }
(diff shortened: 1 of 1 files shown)


7ez:
I'm pretty sure you can use tourney on cutting-edge too, no?

TrueRou:
> I'm pretty sure you can use tourney on cutting-edge too, no?

After testing, sure it can. 

The sent "v" was like `b20250702tourney`, so we might have to join stable and cutting-edge builds on tourney channels checking, as we can't distinguish between them by "v"

arily:
> I have another related question,
> 
> ```python
>     if osu_version.stream is OsuStream.TOURNEY and not (
>         user_info["priv"] & Privileges.DONATOR
>         and user_info["priv"] & Privileges.UNRESTRICTED
>     ):
>         # trying to use tourney client with insufficient privileges.
>         return {
>             "osu_token": "no",
>             "response_body": app.packets.login_reply(
>                 LoginFailureReason.AUTHENTICATION_FAILED,
>             ),
>         }
> ```
> 
> (
> 
> https://github.com/osuAkatsuki/bancho.py/blob/74290910d3ce6e0284453ad9f913ede6a8555fa2/app/api/domains/cho.py#L725
> 
> )
> shouldn't `TOURNEY_MANAGER` be given access to tourney client as well?

https://github.com/ppy-sb/bancho.py/pull/10

PR #718, opened 2025-08-30
Update Dockerfile to remove specific version of build-essential

7ez:
# Describe your changes
removes version pin from the Dockerfile which causes build to fail

https://discord.com/channels/748687781605408908/748687781605408911/1411416892576501862
&
image

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ ENV PYTHONUNBUFFERED=1
 WORKDIR /srv/root
 
 RUN apt update && apt install --no-install-recommends -y \
-    git curl build-essential \
+    git curl build-essential=12.12 \
     && rm -rf /var/lib/apt/lists/*
 
 COPY pyproject.toml poetry.lock ./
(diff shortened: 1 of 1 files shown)


NiceAesth:
would prefer a bump to 12.10 instead, i think this happened because ubuntu-latest is targetting 24.04 lts now instead of 22.04 which had the 12.9 package

7ez:
> would prefer a bump to 12.10 instead, i think this happened because ubuntu-latest is targetting 24.04 lts now instead of 22.04 which had the 12.9 package
> 
> edit from discord: first ubuntu version to be on 12.12 is 25.04 which would be kinda worrying the docker images should not be targetting non lts versions ever 12.10 is current lts

latest python 3.11 image upgraded to debian trixie which appears to not have 12.10 built so theres an issue with that
image

7ez:
me when my system upgrades and breaks my dependencies

NiceAesth:
> me when my system upgrades and breaks my dependencies

real

NiceAesth (approved):

PR #725, opened 2025-10-10
feat: initial bss implementation

### LINKED ISSUE #93: Beatmap submission system
This one is pretty large compared to any other issues, it entails the implementation of the following handlers:
```py
POST /web/osu-osz2-bmsubmit-post.php
POST /web/osu-osz2-bmsubmit-upload.php
GET /web/osu-osz2-bmsubmit-getid.php
GET /web/osu-get-beatmap-topic.php
```
along with writing a program to write osz2 files, and quite a few backend changes.
I've already done a few hundred lines of code of this one, but it will likely not be implemented into master for a while to come.

7mochi:
Ported from Titanic. Closes #93 

# Describe your changes

## Checklist


cmyui:
How far did you get into this? Do you remember how much is working?

7mochi:
> How far did you get into this? Do you remember how much is working?

afaik the core was done, might be missing a few edge cases though

Lekuruu:
Hi, let me jump in to suggest for some changes / fixes, which I applied to Titanic's Beatmap Submission System, that would also affect this pr.

### Shuffled beatmap IDs

The ID assignment in `update_beatmap_metadata` can shuffle beatmap IDs between difficulties, since IDs get resolved one difficulty at a time. The fix is to first pre-assign all IDs that are present in the .osu metadata, and only then hand out the remaining IDs. The `current_bmap_ids + new_bmap_ids` return in `update_beatmaps` has the same ordering problem.
See osuTitanic/deck@f18121f and osuTitanic/deck@6c1c69f.

### `resolve_beatmap_id` logic

Two issues in the current version: it trusts `beatmap.beatmap_id` from the .osu file without checking that it actually belongs to this set. Additionally, `bmap_ids.remove(bmap_object["id"])` raises a ValueError when the ID isn't in the list.
Fixed in osuTitanic/deck@f18121f and osuTitanic/deck@b14c7b9, i.e. only accept an ID if it's in bmap_ids, and remove it from the list once consumed.

### Total length calculation

`last_object - first_object` turned out to be not quite right apparently... I ended up clamping it with `max(..., 0)` after a few bug reports (osuTitanic/deck@d85118d, osuTitanic/deck@fffc868).

### BOM handling when parsing .osu files

`osu_file.decode(errors="ignore")` leaves a `\ufeff` at the start of files saved with a [UTF-8 BOM](https://stackoverflow.com/questions/2223882/whats-the-difference-between-utf-8-and-utf-8-with-bom), which osu stable produces. Use `decode(encoding="utf-8-sig", errors="ignore")` instead (osuTitanic/deck@0fde839).

### File permissions in .osz packages

`ZipInfo` objects created without `external_attr` result in files with no permissions when extracted on unix systems. Setting `zip_info.external_attr = 0o666 << 16` in `create_osz_package` would resolve that (osuTitanic/deck@58cc9d7, osuTitanic/deck@2ec3bbe).

Personally, I'm not too sure about how osu! stable handles this, but I suppose it is better to be safe here. Perhaps we should do a bit more testing on this :shrug: 

### Upload size check happens after reading into memory

`submission_file.file.read()` runs before the 100mb check, so a client can push arbitrarily large bodies into server memory. yikes... that sure didn't go wrong in production ^^
(osuTitanic/deck@57803a3)

Hope this helps. Happy to answer questions about any of these! :D
Oh yeah, it would also be worth to check if `slider` and `osz2` are both up-to-date.

PR #730, opened 2025-10-27
Add redis environment: REDIS_PASS

skchqhdpdy:
# Describe your changes
In the docker-compose.yml file, the Redis section does not have an environment: entry, so the environment variables set in the .env file are not passed into the Docker container. I suspect this is why the redis-server always runs without a password.

I’m not very familiar with using Docker, so I asked ChatGPT, and the response was that the environment: entry is necessary for the password to be applied.

The YouTube link below shows videos I recorded myself using redis-cli.exe to connect.
https://youtu.be/72fQFjOm_I8
The first demonstration shows connecting when the environment: option is missing.
The second demonstration shows connecting when the option is present.
The third demonstration shows attempting to connect again when the option is missing.

## Checklist

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -32,6 +32,8 @@ services:
       - test-redis-data:/data
     networks:
       - test-network
+    environment:
+      - REDIS_PASS=${REDIS_PASS}
     command:
       - sh
       - -c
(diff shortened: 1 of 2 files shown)


NiceAesth:
a bit confused as to why this would work

https://github.com/redis/docker-library-redis/blob/c5846c13383c8b284897ff171e0c946868ed4c8c/debian/docker-entrypoint.sh @7mochi any input?

PR #765, opened 2026-04-03
replace assert with explicit check in comment validation

### LINKED ISSUE #760: Comment handler uses assert for input validation
The comment POST handler checks required params (`target`, `start_time`, `comment`) with bare `assert` statements. Running python with `-O` strips all asserts from bytecode, so these validations just vanish.

`app/api/domains/osu.py`, lines 1471-1473

raya-ac:
Fixes #760

Swapped the bare `assert` statements to a regular conditional that returns 400. Asserts get stripped under `python -O` which would remove these checks entirely.

### DIFF
--- a/app/api/domains/osu.py
+++ b/app/api/domains/osu.py
@@ -1468,9 +1468,8 @@ async def osuComment(
         # client is submitting a new comment
 
         # validate all required params are provided
-        assert target is not None
-        assert start_time is not None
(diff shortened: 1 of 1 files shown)


NiceAesth:
do not run the client with python -O this isn't a real issue

there are probably hundreds of other places asserts are used, changing this one spot does not achieve much

PR #816, opened 2026-07-06
Compose application dependencies explicitly at startup

cmyui:
## Summary

- Builds one explicit `Application` graph during lifespan startup: runtime resources, repositories, isolated session/cache state, domain services, command router, packet router, and supervised background tasks.
- Replaces the mutable `app.state` service/session/cache/packet registries and the legacy repository locator. FastAPI keeps only the root application at its framework boundary.
- Gives each application instance its own command and packet registries. Command dependencies are bound with typed registration, so wiring errors are checked by mypy and tests can register local commands without touching process-global state.
- Moves persistence, cache, network, moderation, relationship, beatmap, score, and player-session work out of active-record-style entities and into injected services.
- Centralizes fire-and-forget task ownership and guarantees resource cleanup when startup or composition fails partway through.
- Updates the recalc tool and test fixtures to compose the dependencies they use directly; the test suite no longer monkeypatches application globals.

## Behavior and regression coverage

- Preserves all 46 packet registrations and all 8 restricted-player packet permissions.
- Preserves transaction participation by sharing one database wrapper across repositories.
- Preserves restricted-player rank behavior (rank `0` with no Redis lookup) and the single global-rank lookup per unrestricted mode.
- Makes friend/block type changes one atomic upsert and updates in-memory caches only after persistence succeeds.
- Publishes a newly created match only after its host joins successfully, preventing hostless matches and leaked match slots.
- Keeps background latest-activity writes non-blocking, cleans up partially opened startup resources, and represents database-hydrated offline players with an empty token.

## Testing

- Full MySQL/Redis suite: **472 passed**
- Coverage: **74%** (required: 70%)
- Strict mypy: **no issues in 138 source files**
- Pre-commit: **all hooks passed**


cmyui:
This is the most insane PR in the history of all PRs

app/api/domains/cho.py
```diff
@@ -1033,18 +1228,25 @@ async def handle_osu_login_request(
...
     )
 
-    player.update_latest_activity_soon()
+    player_data_service.schedule_latest_activity_update(player)
 
     return {"osu_token": player.token, "response_body": bytes(data)}
 
 
-@register(ClientPackets.START_SPECTATING)
 class StartSpectating(BasePacket):
```
cmyui, line 1236: @7mochi see how my broadcast still stayed here, packet stuff stayed in handler while service handled domain logic. probably still can be cleaner here in the handler, but the boundary is intentional

cmyui: don't read into this pr too much other than this area, it's exploratory

NiceAesth (approved):
Image
right....

```py
if diffstat.linecount > 1000:
    await pr.respond("LGTM!")
```

PR #818, opened 2026-07-17
fix: remove player from session indexes before invalidating token on logout

MitzaBobitza:
# Describe your changes

Every player logout currently raises `KeyError: ''` in the ASGI logs:

```
  File "/srv/root/app/objects/player.py", line 400, in logout
    app.state.sessions.players.remove(self)
  File "/srv/root/app/objects/collections.py", line 261, in remove
    del self._by_token[player.token]
KeyError: ''
```

`Player.logout()` blanks `self.token = ""` **before** calling `players.remove(self)`, but since #747 the `Players` collection keeps a `_by_token` index keyed by the login token — so the removal looks up the already-blanked token and raises.

The impact goes beyond log noise: the `KeyError` aborts `Players.remove()` after `super().remove(player)` but before the index cleanups, so `_by_token`/`_by_id`/`_by_name` keep stale entries pointing at the logged-out player (e.g. `players.get(id=...)` still resolves them), and the logout packet is never broadcast to other online players.

This PR reorders `logout()` to remove the player from the collection first and invalidate the token after. The reorder is safe: `logout()` is fully synchronous (runs atomically on the event loop), and none of the intermediate steps (`leave_match`, `remove_spectator`, `leave_channel`) read `self.token` / `is_online`.

Also adds a regression test that fails with `KeyError` on current master and passes with the fix.

## Related Issues / Projects

Regression introduced in #747.

## Checklist

New regression test + full unit suite (334 tests) pass, mypy clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


cmyui:
Thank you for the fix!

PR #824, opened 2026-07-17
Raise coverage floor to 70%

cmyui:
## What changed

Raise the CI statement-coverage floor from 45% to 70%.

## Why

Master currently reports >70% coverage, so the previous 45% threshold allowed a large regression before CI would fail. A 70% floor meaningfully protects the existing suite while retaining a small margin for incidental coverage variation.

This is intentionally separate from the new coverage PRs so the CI policy change can be reviewed independently.

## Validation

- `make test` (388 passed, 68% coverage with the new 65% floor)
- pre-commit hooks for `scripts/run-tests.sh`


scripts/run-tests.sh
```diff
@@ -57,5 +57,5 @@ execDBStatement "source /srv/root/migrations/base.sql"
 # Run tests
 echo "Running tests..."
 coverage run -m pytest -vv -s tests/
-coverage report --show-missing --fail-under=45
+coverage report --show-missing --fail-under=65
```
cmyui:
```suggestion
coverage report --show-missing --fail-under=70
```

